PR #6, opened 2023-02-22
Check default locations if no --config

thewizzy:
Uses dirs crate to provide a cross platform approach to finding configuration directories.
Linux: ${XDG_CONFIG_HOME}
Windows: {FOLDERID_RoamingAppData}
MacOS: $HOME/Library/Application Support

Falls back to `/etc/leaktk/config.toml' on unix systems

Fall back to the Default::default()


Cargo.toml
```diff
@@ -18,3 +18,4 @@ ring = "0.16"
 log = { version = "0.4.17", features = ["std", "serde"] }
 time = "0.3.17"
 uuid = { version = "1.2.2", features = ["v4", "fast-rng"] }
+dirs = "4.0"
```
bplaxco, line 21: @thewizzy could have you add in that newline at the end of the file?

src/main.rs
```diff
@@ -16,7 +16,7 @@ use crate::scanner::Scanner;
 fn main() -> Result<(), Error> {
     let config = match parser::args().config {
         Some(config_path) => Config::load(&config_path)?,
```
bplaxco, line 18: @thewizzy what are your thoughts on tweaking `Config::load` to take an optional path (replacing this) and tweaking the logic to look like:

```
If arg provided path:
  use it with out seeing if it exists
else if one of the default paths exist:
   use it
else:
   return Ok(Default::default()) 
```

Then we could do something like:

```rust
let config = Config::load(parser::args().config)?
```

thewizzy: Added this recommendation. 

I kept `load_file` to keep the loading logic separate from the where logic.

PR #7, opened 2023-02-23
Change workdir default to ${XDG_CACHE_HOME}/leaktk

thewizzy:
The cache directory is generally longer lived than system temporary directories. The behavior of env::temp_dir() is that it will land on $TMPDIR then /tmp on unix.

It is unlikely that dirs::cache_dir will return none on a supported OS but just in case reverting to env::temp_dir().

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,7 +3,10 @@ use dirs;
 use serde::Deserialize;
 use std::env;
 use std::fs;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
+
+#[cfg(any(target_os = "linux", target_os = "macos"))]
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -3,7 +3,10 @@ use dirs;
 use serde::Deserialize;
 use std::env;
 use std::fs;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
+
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+use std::path::Path;
```
thewizzy, line 9: Please note that this fixes a warning of unused import when compiling not on linux || macos.

PR #8, opened 2023-03-02
Better Error Handling

bplaxco:
Not quite ready yet, still learning my way around, but just want to see if this is heading in the right direction. Feedback is welcome.


src/errors.rs
```diff
@@ -1,35 +1,79 @@
...
+    #[error("Could Not Save Gitleaks")]
+    CouldNotSaveGitleaks(#[source] std::io::Error),
+
+    #[error("Could Not Remove Invalid Gitleaks")]
+    CouldNotRemoveInvalidGitleaks(#[source] std::io::Error),
+
+    #[error("Could Not Parse Results")]
+    CouldNotParseResults(#[from] serde_json::Error),
+
+    #[error("Could Not Rarse Scan")]
```
thewizzy, line 41:
```suggestion
    #[error("Could Not Complete Scan")]
```

bplaxco:
Goals:

* Implement custom errors that implement `std::error::Error`
* Remove unwraps and expects
* Have user friendly errors returned in the scan results
* Have more detailed errors in the log output
* Sort the imports into sections going from global to local
* Format the error strings to be lower case with no ending punctuation
* Leverage `From` where I could to reduce `map_err` calls.
* Misc refactors

PR #9, opened 2023-03-16
Support allowlists in a repo's .gitleaks.toml

bplaxco:


bplaxco:
TODO:
* Add `results_dir` attr tow the workspace for [this PR](https://github.com/leaktk/scanner/pull/11)
* Fix the bug[1] I introduced for local scans

[1] For a Workspace the root dir should still point at the ephemeral scan dir but:

* the repo_dir should point at the local repo dir
* clean should remove the workspace root dir but not the repo dir

So workspace will probably need some concept of should I delete the repo dir or something like that (which might come in handy for a future feature for preserving evidence).

When this is done, note it in https://github.com/leaktk/scanner/pull/11

PR #12, opened 2023-04-01
Clippy fixes

thewizzy:
This branch is to resolve Clippy issues.

There are no Clippy issues left on main.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -182,7 +182,7 @@ impl Config {
     /// Load the config from a provided file path or fall back on defaults
     pub fn load(path: Option<String>) -> Result<Config, ConfigError> {
         if let Some(path) = path {
-            return Config::load_file(&Path::new(&path));
+            return Config::load_file(Path::new(&path));
         }
 
         if let Some(config_dir) = dirs::config_dir() {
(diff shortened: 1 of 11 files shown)


src/config.rs
```diff
@@ -159,19 +159,19 @@ pub struct Config {
 
 impl Config {
     /// Load the config from a `&str`
-    pub fn from_str(raw: &str) -> Result<Config, ConfigError> {
+    pub fn load_from_str(raw: &str) -> Result<Config, ConfigError> {
```
bplaxco, line 162: @thewizzy thoughts on implementing from_str on it via [the trait](https://doc.rust-lang.org/std/str/trait.FromStr.html)? Do you think that'd be a lot of extra work?

bplaxco (commented):
Dang, I love the feedback that clippy gives :ok_hand:

src/config.rs
thewizzy: I don't think it would be considerable work. This one did give me pause for a sec, I actually started implementing `from_str` trait but it didn't feel right. Happy to take your preference.
It is really just wrapping that function in `impl FromStr for Config `

thewizzy: There is a new commit moving to the `from_str` trait. Take your pick.

PR #13, opened 2023-05-03
Auth header support

bplaxco:


examples/config.toml
```diff
@@ -21,3 +21,6 @@
 # # The URL to a pattern server
 # # The path "/patterns/{scanner}/{version}" will be appended to this URL
 # url = "https://raw.githubusercontent.com/leaktk/patterns/main/target"
+# # If specified, the contents of the file are sent as an Authorization Bearer
+# # token, else an Authorization header is not sent.
+# # auth_token_file_path = "/home/user/.config/leaktk/auth.jwt"
```
thewizzy, line 26:
```suggestion
# auth_token_file_path = "/home/user/.config/leaktk/auth.jwt"
```
Small style change

src/scanner/patterns.rs
```diff
@@ -71,8 +86,30 @@ impl Patterns {
                     .ok_or(PatternsError::PathHasNoParent)?,
             )?;
 
+            let mut headers = header::HeaderMap::new();
+
+            if let Some(path) = &self.server.auth_token_file_path {
+                let auth_token =
+                    fs::read_to_string(path).map_err(PatternsError::CouldNotOpenAuthTokenFile)?;
```
thewizzy, line 93: For curiosity why did you select a separate file for this? Was it for easier secrets management as the config is not particularly sensitive and you can keep them in separate places?

thewizzy:
The program continues running even if there is no pattern file and the auth is malformed. Would it be best to bail out on a malformed auth token? 
```
{"id":"bca55c27-a0e1-42b6-91b4-1911017dd2a8","kind":"git","target":"https://github.com/leaktk/fake-leaks.git"}
{"level":"INFO","message":"refreshing patterns","time":"2023-05-04 4:19:25.234666749 +00:00:00"}
{"level":"ERROR","message":"auth token is invalid: failed to parse header value","time":"2023-05-04 4:19:25.234936659 +00:00:00"}
{"id":"a5446a02-6638-4583-8538-0b6adbf80023","request":{"id":"bca55c27-a0e1-42b6-91b4-1911017dd2a8"},"error":"auth token is invalid: failed to parse header value","results":[]}
```

bplaxco:
Adding a couple of additional todos that I forgot:

* Support setting the token via a `LEAKTK_PATTERN_SERVER_AUTH_TOKEN` env var
* Support setting the URL via a `LEAKTK_PATTERN_SERVER_URL` env var

src/scanner/patterns.rs
bplaxco: Yeah it was so that tools that implement a login sub-command could just write to the file once they get a token rather than having to update the config. That being said, one additional thing I need to add is:

If `~/.config/leaktk/pattern-server-auth-token` exists and `auth_token_file_path` isn't defined, use that file by default.

examples/config.toml
bplaxco: Addressed :ok_hand:

src/scanner/patterns.rs
bplaxco: Updated so that the config if set and contains the actual auth token (instead of a path) it wins out but if it's not set you can fall back on LEAKTK_PATTERN_SERVER_AUTH_TOKEN then ~/.config/leaktk/pattern-server-auth-token.

bplaxco:
> The program continues running even if there is no pattern file and the auth is malformed. Would it be best to bail out on a malformed auth token?
> 
> ```
> {"id":"bca55c27-a0e1-42b6-91b4-1911017dd2a8","kind":"git","target":"https://github.com/leaktk/fake-leaks.git"}
> {"level":"INFO","message":"refreshing patterns","time":"2023-05-04 4:19:25.234666749 +00:00:00"}
> {"level":"ERROR","message":"auth token is invalid: failed to parse header value","time":"2023-05-04 4:19:25.234936659 +00:00:00"}
> {"id":"a5446a02-6638-4583-8538-0b6adbf80023","request":{"id":"bca55c27-a0e1-42b6-91b4-1911017dd2a8"},"error":"auth token is invalid: failed to parse header value","results":[]}
> ```

At first I was on the fence, but after typing out a long drawn out response and thoughts on the options, I erased it all and I agree with you. 

Since there's a few spots that I might need to tweak, I added a `Identify non-recoverable errors and have them log and exit` so I can think through error handling then and figure out how I want to log and exit in those cases.

(Note to self, see if there's a way to customize how panic! behalves and maybe log using error! and exit)

With the goal to do that new task prob this Wed, how does the rest of the code look here?

One other thing that feels odd in the implementation is config values have greater precedence than the env variables. If you agree that feels off I'll probably tweak that this Wed as well, but that ended up being easier to implement while snagging some coffee lol

thewizzy:
LGTM

PR #14, opened 2023-12-01
Start parsing config

bplaxco:
There will be more done to parse the config but this is just an initial port of the config objects that seemed relevant


pkg/config/config.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// PatternServer provides pattern server configuration settings for the scanner
+	PatternServer struct {
+		URL       string `toml:"url"`
+		AuthToken string `toml:"auth_token"`
+	}
+)
+
+func xdgDir(defaultName, envVar string) string {
```
thewizzy, line 51: Would you consider using something like https://github.com/adrg/xdg to do this cross platform? This ones dependencies are already in here anyway.

thewizzy:
Looks good to me, you can merge it and we can consider ardg/xdg in the future.

abutcher (commented):
Looks good! Added some `assert` suggestions for expected errs/notnils in the tests.

PR #15, opened 2023-12-01
Parse CLI args and load the config

bplaxco:
This should be done after https://github.com/leaktk/scanner/pull/14


thewizzy (approved):

abutcher (approved):
Looks great!

PR #18, opened 2024-06-03
Most of the go rewrite

bplaxco:
I'm going ahead and kicking off this draft MR to get started on this since it's pretty big. I think what's left is checking off some todos, updating the README, more tests, fixing up the patterns code so it looks at the timeouts before fetching again like it ought to.

It might be easiest to just clone the new branch and pick through it rather than looking at the diffs.

There's also no rush on this since I'm still picking at it.


pkg/resource/json.go
```diff
@@ -0,0 +1,183 @@
...
+// that have versions
+func (r *JSONData) Since() string {
+	return ""
+}
+
+// ReadFile provides a way to access values in the JSON data
+func (r *JSONData) ReadFile(path string) ([]byte, error) {
+	// Build out the path components
+	var components []string
+	for _, component := range strings.Split(path, string(filepath.Separator)) {
```
thewizzy, line 104: This is problematic for cross platform. 
This is all in memory, right? If that is the case should we standardize on the representation within the json being "unix"?

pkg/scanner/gitleaks_test.go
```diff
@@ -0,0 +1,59 @@
...
+	configPath := filepath.Join(tempDir, "gitleaks.toml")
+	err = os.WriteFile(configPath, []byte(mockGitleaksTestConfig), 0644)
+	assert.NoError(t, err)
+
+	// Configured patterns
+	patterns := NewPatterns(
+		&config.Patterns{
+			Gitleaks: config.Gitleaks{
+				ConfigPath: configPath,
+			},
```
thewizzy, line 36: This is where the issue I shared with you last week stems from. 

I suspect you have a config.toml somewhere it is expecting it, I did not. So i did not have the ExpiredAfter (or any others settings configured). Adding the following resolved it.
 
```suggestion
			},
			ExpiredAfter: 60 * 60 * 24 * 7,
```

bplaxco: @thewizzy I went ahead and used `config.DefaultConfig()` here and in some other places. That should set that as well as some other relevant values. Does that also resolve the issue for you?

pkg/resource/json.go
bplaxco: I want `ReadFile` to be consistent across all the different resource types and work with the output of [filepath.Join](https://pkg.go.dev/path/filepath#Join). Since it joins elements with an OS specific [Separator](https://pkg.go.dev/path/filepath#Separator) I wanted to make sure they lined up.

So in theory this should make it less problematic cross platform, right?

pkg/scanner/gitleaks_test.go
thewizzy: Default config works well.

pkg/resource/json.go
thewizzy: Sounds reasonable. 
We can use `filepath.join` to create `path` entries in the tests. I've sent a diff separately.

pkg/logger/logger.go
```diff
@@ -0,0 +1,205 @@
...
+	NOTSET LogLevel = 0
+	// DEBUG will enable these logs and higer
+	DEBUG LogLevel = 10
+	// INFO will enable these logs and higer
+	INFO LogLevel = 20
+	// WARNING will enable these logs and higer
+	WARNING LogLevel = 30
+	// ERROR will enable these logs and higer
+	ERROR LogLevel = 40
+	// CRITICAL will enable these logs and higer
```
abutcher, line 34:
```suggestion
	// DEBUG will enable these logs and higher
	DEBUG LogLevel = 10
	// INFO will enable these logs and higher
	INFO LogLevel = 20
	// WARNING will enable these logs and higher
	WARNING LogLevel = 30
	// ERROR will enable these logs and higher
	ERROR LogLevel = 40
	// CRITICAL will enable these logs and higher
```

pkg/resource/git.go
```diff
@@ -0,0 +1,218 @@
...
+	return r.options.Since
+}
+
+// ReadFile provides a way to get files out of the repo
+func (r *GitRepo) ReadFile(path string) ([]byte, error) {
+	object := fmt.Sprintf("HEAD:%s", filepath.Clean(path))
+	return exec.Command("git", "-C", r.ClonePath(), "show", object).Output() // #nosec G204
+}
+
+// GitDirPath returns the path to the git dir so that other things don'g need
```
abutcher, line 161:
```suggestion
// GitDirPath returns the path to the git dir so that other things don't need
```

abutcher (commented):
Some typos from a walk-through :smile_cat:

PR #19, opened 2024-06-26
Files scan

bplaxco:
Support scanning files and directories.


bplaxco:
Going ahead and pulling this in so I can start integrating it into the pattern testing

PR #22, opened 2024-07-08
Fixing linting issues

thewizzy:


pkg/resource/files.go
```diff
@@ -93,14 +93,14 @@ func (r *Files) Walk(fn WalkFunc) error {
 	}
 
 	return filepath.WalkDir(r.path, func(path string, d iofs.DirEntry, err error) error {
-		relPath, err := filepath.Rel(r.path, path)
 		if err != nil {
-			logger.Error("could generate relative path: path=%q error=%q", path, err)
+			logger.Error("could not walk path: path=%q error=%q", path, err)
```
thewizzy, line 97: This error was further down and wouldn't have really been reached. I suspect it was mistakenly there and was originally further up to catch and consume the WalkDir func error so it can be logged and not passed up to cancel all the walking.

bplaxco (approved):
LGTM!

PR #26, opened 2024-07-17
Check branch before cloning

bplaxco:
If the branch doesn't exist on the remote, expand the scan scope to all branches.


thewizzy:
Looks good to me. 

I would recommend bumping Go to version 1.21 (in go.mod and in .github/workflows/testing.yml). 1.21 is in pretty minimal these days.

PR #34, opened 2024-08-23
Fix panic on invalid remote gitleaks.toml

thewizzy:
If the regex inside a gitleak toml file is invalid it will cause a panic.

This code adds a `recover` and allows the main process and other routines to continue. It bails out of the rest of the ParseGitleaksConfig, hence having to add a check for a nil config.

### DIFF
--- a/pkg/scanner/patterns.go
+++ b/pkg/scanner/patterns.go
@@ -151,16 +151,16 @@ func invalidConfig(cfg *gitleaksconfig.Config) bool {
 }
 
 // ParseGitleaksConfig takes a gitleaks config string and returns a config object
-func ParseGitleaksConfig(rawConfig string) (*gitleaksconfig.Config, error) {
+func ParseGitleaksConfig(rawConfig string) (glc *gitleaksconfig.Config, err error) {
 	var vc gitleaksconfig.ViperConfig
 
 	defer func() {
-		if recover() != nil {
-			logger.Error("Gitleaks config is invalid. Recovering")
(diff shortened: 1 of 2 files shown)


pkg/scanner/patterns.go
```diff
@@ -154,6 +154,12 @@ func invalidConfig(cfg *gitleaksconfig.Config) bool {
 func ParseGitleaksConfig(rawConfig string) (*gitleaksconfig.Config, error) {
 	var vc gitleaksconfig.ViperConfig
 
+	defer func() {
+		if recover() != nil {
+			logger.Error("Gitleaks config is invalid. Recovering")
+		}
+	}()
```
bplaxco, line 161:
```suggestion
func ParseGitleaksConfig(rawConfig string) (*gitleaksconfig.Config, err error) {
	var vc gitleaksconfig.ViperConfig

	defer func() {
		if r := recover(); r != nil {
			err = fmt.Errof("gitleaks config is invalid: error=%q", r)
		}
	}()
```

I *think* that should avoid needing to do the clonedConfig nill check. I did a test script and it looked like it should work :+1: Just basically using the named return values. It also gets the cause of the error in the returned response :+1: 

If that works and and we can avoid `|| clonedConfig == nil` then feel free to merge that change on in. If that doesn't work and we still have to do the nil check for some reason, then I'm also cool with what we have here.

thewizzy: Applied separately as it had a couple of other touch points.

PR #37, opened 2024-08-29
Bundle major errors into Response

thewizzy:
This feature is a must, it makes it much easier for other systems consuming the output to tie together errors with responses.

The implementation 🤷 . This only covers major errors, only the ones that are currently tied to request id's already. We could add further to error holding to the resource. This could help if we want to know whether the local .gitleaks.toml files failed or other similar errors.

### DIFF
--- a/pkg/scanner/request.go
+++ b/pkg/scanner/request.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
 	"github.com/leaktk/scanner/pkg/logger"
 	"github.com/leaktk/scanner/pkg/resource"
 )
@@ -14,6 +13,7 @@ type Request struct {
 	ID string
 	// Thing to scan (e.g. URL, snippet of text, etc)
 	Resource resource.Resource
+	Errors   []LeakTKError
 }
 
 // UnmarshalJSON sets r to a copy of data
--- a/pkg/scanner/scanner.go
+++ b/pkg/scanner/scanner.go
@@ -1,6 +1,7 @@
 package scanner
 
 import (
+	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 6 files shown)


pkg/scanner/request.go
```diff
@@ -14,6 +14,13 @@ type Request struct {
...
 	// Thing to scan (e.g. URL, snippet of text, etc)
 	Resource resource.Resource
+	Errors   []RequestError
+}
+
+type RequestError struct {
+	Fatal   bool
+	Context string
+	Message string
 }
```
bplaxco, line 24: What if we implement the [error interface](https://pkg.go.dev/builtin#error) and swap `Context` with a `Kind` field with strings like `CloneError`, `ResourceCleanupError`, etc?

Also instead of having a separate RequestError and ResponseError which require some mapping at the end, we could do a single `ScannerError` or even a more generic `LeakTKError` if we want to reuse this pattern in other bits of code.

Rationale:

- With a single error type we wouldn't need to do any mapping at the end
- We can change `Errors` type to `[]error` and tack what ever error type we want in it.
- We could (not in this MR) fix the multiple levels of escape characters issue with a custom error type by having it put them in a field on `LeakTKError` like `From []error` or something that we just collect the errors that bubbled up and caused this one. Then it'd render as just a nice clean JSON structure.
- If we implement the error interface we can start replacing other error creation in the code with this new error type without changing any of the code really. 
- Using `Kind` instead of `Context` helps communicate that this field is stable enough for things down the line to match on it and make decisions on it, and any human readable details would land in the Message field.

Thoughts?

bplaxco:
I like it :+1: 

I added one comment that I think would help out with some other issues as well and make it a bit more flexible in the code.

pkg/scanner/request.go
thewizzy: Pushed updated code. This should meet the above comment and our slack chat. Its extensible to do dots 2-4 from above at a later stage.

pkg/scanner/scanner.go
```diff
@@ -136,21 +142,39 @@ func (s *Scanner) listenForScanRequests() {
...
 			logger.Error("skipping scan due to missing clone path: request_id=%q", request.ID)
+			request.Errors = append(request.Errors, LeakTKError{
+				Fatal:   true,
+				Code:    CloneError,
+				Message: fmt.Sprintf("missing clone path: request_id=%q (%s)", request.ID, reqResource.ClonePath()),
+			})
+		}
+		for _, e := range request.Errors {
+			fmt.Println(e)
 		}
```
bplaxco, line 173:
```suggestion
```

I don't think we want to include the errors in stdout here since they'll be passed along.

bplaxco:
Added one small comment on something that looks like it might be a leftover. If that can be removed then feel free to merge on it.

PR #40, opened 2024-09-04
Fix intermitent test failures

thewizzy:
The test fails as there are a few "fake" and "FAKE" in the repo.  
Depending on which one it finds first will determine if it passes.  

Resolved by converting to lower case for all results.


bplaxco:
LGTM

PR #45, opened 2024-09-18
Add completions folder

thewizzy:
Followed how podman handles their completions in their repo and also for their RPM specfile. 

Once this and #38 are merged we can cut v0.0.2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,12 +5,27 @@ BUILD_META += -X=github.com/leaktk/scanner/cmd.Version=$(VERSION)
 BUILD_META += -X=github.com/leaktk/scanner/cmd.Commit=$(COMMIT)
 PREFIX ?= /usr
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -36,9 +51,25 @@ test: format gosec golint
...
+
 security-report:
 	trivy fs .
 
 update:
 	go get -u ./...
 	go mod tidy
+
+.PHONY: validate.completions
+validate.completions: SHELL:=/usr/bin/env bash
```
bplaxco, line 71: Interesting. I didn't know you could specify the target twice like this. What does it do?

thewizzy: Nor did I. It is in the Podman [Makefile](https://github.com/containers/podman/blob/main/Makefile#L616)

bplaxco (approved):
LGTM!

PR #46, opened 2024-09-18
Calculate & log the sha256 hash of patterns on update

thewizzy:
For our use case we would like to know when the patterns have changed so that we can extend our look-back and re-scan some repositories that have already been scanned with the previous patterns. 
This change calculates the sha256 hash of the patterns on loading and after each refresh - sending this out as part of a logger.Info message.


bplaxco (approved):
LGTM

PR #47, opened 2024-09-19
Quieten down pattern hash update logs

thewizzy:
The previous one worked but it sends a log message every single scan. Modified it so that the hash value is only updated when it is changed and we are only notified when it changes.

Sorry for the noise, it works - but its a log entry for every scan. I didn't realize we didn't reuse patterns and call this every time.


bplaxco (approved):
LGTM

PR #49, opened 2024-09-23
Fix git log failing on invalid objects

thewizzy:
When doing a shallow commit there is a list of hashes in the shallow file. Sometimes these are invalid. We want to not include the list of hashes in this list otherwise we will end up including secrets included before the shallow commit. This is more likely to happen on busy repositories.

The change here allows us to skip and invalid hashes (likely commits in branches that have since been squashed/deleted).

Added a fix for a spelling error as well.


thewizzy:
Merging myself on agreement with bplaxco

PR #50, opened 2024-09-24
Log when a URL fetch in JSONData fails rather than failing the whole scan

thewizzy:
The change is small,  instead of passing the error up, which would cause the clone to fail and the JSONData to not be scanned. This logs that a URL was unable to be fetched, passing back no error and allowing scanning to continue.

This is a short term solution until we improve our general error handling. Allowing fatal/nonfatal errors to be floated up to the application.

Added a few items to the to do list, including the aforementioned error handling.


thewizzy:
Merging per agreement with bplaxco. This is a temporary fix that allows the intended behavior, whilst we improve our overall fatal/nonfatal error handling.

PR #51, opened 2024-09-24
Only return request_id, not the whole request.

thewizzy:
The whole request can be quite sizeable, especially with JSONData. The requester should know their unique request ID so can link using only that.


bplaxco (approved):
LGTM

PR #52, opened 2024-09-25
Include and run goimports with make format

bplaxco:


thewizzy:
Maybe we can add goimports checks to golangci-lint ?

PR #55, opened 2024-09-30
Use a flexible priority queue

bplaxco:
This will keep stdin from blocking since the heap in the queue can expand as much as it needs.

Also if we add a global `priority` option to all the requests, we can feed it in.

There seems to be a little issue with the timing in the tests that's messing with the order a little. I need to fix that and test some things, but what do y'all think?


thewizzy (approved):
Looks good. 

Would file/folder naming make more sense if it matched the structs. 
Like messageheap.go or priorityqueue.go ?

What caused the bump to version 1.22?

bplaxco:
I'll update the file names 👌

I was seeing a warning, but let me explore that more and see if I can bump it back down to 1.22 without it. I was doing the queue a different way when I was originally getting it.

bplaxco:
Go version set back to 1.21. There was a small change in how for...range loops work between 1.21 and 1.22 that was throwing me for a loop lol (badumtss)

bplaxco:
It looks like this actually is a tad faster than what we had before.

abutcher (approved):
Looks good to me!

PR #58, opened 2024-10-09
Bump gitleaks to 8.20.1

thewizzy:
- The validation of config requires the minimal config to have an ID now. Updated tests
 - As of 8.20 the go version required is 1.22


bplaxco (approved):
LGTM! :shipit: :

PR #74, opened 2024-11-07
Local scans

bplaxco:
**Very much a WIP**

- Have resources have a concept of if they're local or to skip cloning
- Support local git scans
- Support scanning unstaged and staged items


README.md
```diff
@@ -84,6 +86,24 @@ Notes about the formats below:
...
+
+```json
+{
+  "id": "db0c21127a6a849fdf8eeae65d753275f3a26a33649171fa34af458030744999",
+  "kind": "GitRepo",
+  "resource": "./fake-leaks.git",
+  "options": {
+    "branch": "main",
+    "depth": 1,
+    "local": true,
```
thewizzy, line 104: As a user of the command line, I would find it quite frustrating to write json for something like this. 
`leaktk-scanner scan --resource '/path/to/resource' --options '{"local":true}'` 

I think we should support the option, but also support just putting a valid filesystem path in the --resource

The following would cover validating a local git repo
`if fs.PathExists(r.cloneURL) && (fs.PathExists(filepath.Join(r.cloneURL, ".git")) || fs.FileExists(filepath.Join(r.cloneURL, ".git")))` works for both git and git worktrees. We could also add in some validation so that it has to be a valid URI (though the current error is basicallycannot clone /path/to/git/repo

bplaxco: For that it would auto detect. I think in the comment I linked it was only in the listen requests that you have to explicitly set it. Just because listen probably won't be using it very often and there's no reason to do a whole lot of file checks.

thewizzy: I've rebased it to the `update-gitleaks` branch now. 
I'll figure out a way for it to autodetect for a `scan` and not for listen.

thewizzy: We should, regardless of whether it is scan or listen, automatically detect whether it is a remote or local resource. This however opens a potential vulnerability. Given local is less likely, we can check for valid remote URIs for each resource kind first, before doing file system calls.

We should gate allowing `local` on listen behind a configuration value - so that it can be systematically blocked. 
It would be good to seamlessly auto detect local vs remote for Git and ContainerImage.

Thoughts?

thewizzy: I've added it so that for `scan` command it will set the option based on what it determines. This is in line with the documentation.

I think we should add an `allow_local` setting in the config file. It would be good to block local scans on listen at that level. Maybe I am a bit too paranoid?

bplaxco: I definitely like the idea of being able to disable/enable local scanning. Heh, I'm also paranoid about listen (though there's a lot of additional hardening it would need to go through before it should be allowed fully untrusted input). Also there's a slight performance gain with turning the check completely off.

Yeah I think I like both defaulting to detecting local resources and having a scanner level `allow_local` flag that disables any form of local scanning.

```go
if r.IsLocal() && !s.allowLocal {
    // log an error and skip the scan
}
```

That would also stop things like Files scans, etc.

thewizzy: Ive added this and pushed. But i need to work on it a bit. When local is allowed on remote, it will happily scan it even if local isn't explicitly set as an option - `git clone 'folder' 'folder'` works as you'd expect.

thewizzy: The following commit attempts to allow `git clone 'folder' 'folder'` but also block it if `allow_local` is false.

https://github.com/leaktk/scanner/pull/74/commits/313ab9390e87e6d8123a43360f149445d0382636

pkg/scanner/patterns.go
```diff
@@ -155,7 +155,11 @@ func (p *Patterns) updateGitleaksConfigHash(hash [32]byte) {
 }
 
 func invalidConfig(cfg *gitleaksconfig.Config) bool {
```
thewizzy, line 157: @bplaxco We can probably remove this now as we can just append the AllowLists to the scanner (gitleaks.go:89). 

We call `Translate()` which then calls `parseAllowlist`:
https://github.com/gitleaks/gitleaks/blob/6f967cad68d7ce015f45f4545dca2ec27c34e906/config/config.go#L144

`parseAllowList` then calls `Validate`
https://github.com/gitleaks/gitleaks/blob/master/config/config.go#L266

We return the error from Translate which would return an empty config. 
Thoughts?

bplaxco: I think I'll leave it in since that rules check is light and it's also a nice sanity check that the allowlist is doing something useful :ok_hand:

PR #76, opened 2024-11-12
Support gitleaks max decode depth

bplaxco:


pkg/scanner/gitleaks.go
```diff
@@ -25,13 +25,15 @@ const (
...
 // Gitleaks wraps gitleaks as a scanner backend
 type Gitleaks struct {
-	patterns *Patterns
+	patterns       *Patterns
+	maxDecodeDepth uint16
 }
 
 // NewGitleaks returns a configured gitleaks backend instance
-func NewGitleaks(patterns *Patterns) *Gitleaks {
+func NewGitleaks(maxDecodeDepth uint16, patterns *Patterns) *Gitleaks {
```
thewizzy, line 33: Should we swap the parameter order or the order of the fields in the struct?

PR #87, opened 2025-01-29
Give group same file permissions as owner

bplaxco:
This should help in containerized contexts where uid might not be consistent.


bplaxco:
Decided to do this a different way.

PR #93, opened 2025-02-10
Add --format output options

### LINKED ISSUE #84: RFE: Pretty output option for scan
Nice to have:

```
leaktk-scanner scan --format={pretty,yaml,json} ...
```

or something similar so that the output is more human readable for ad-hoc scans.

The default will stay `json`. But other kinds that might make sense are:

* csv - just flatten the field names for nested values (see the example at the end for field names)
* yaml - just format as yaml instead of json
* toml - just format as toml instead of json
* pretty or human or something similar. Maybe with a format like this?

```
ID:                       d2f80dc543822e4a319542914864dd7cd961cf8ef2e1d4bfe68b06871474fe5b
REQUEST.ID:               46a96aa1a87283d055b8d1e297b8fde379255d09058b008b4346384dea2bb470
RESULT.ID:                fde57c9bd7a0f2969dce4f7d526dfe550fa031e5711ed624c85d43e3e11a993d
RESULT.KIND:              GitCommit
RESULT.SECRET:            ...
RESULT.MATCH:             ...
RESULT.ENTROPY:           5.0 
RESULT.DATE:              ...
RESULT.RULE.ID:           cdcf75b34688c4e0d60fa7a6e76fb4ca38f979b3278c4bfed23ebb1366790e22
RESULT.RULE.DESCRIPTION:  Generic Secret
RESULT.TAGS:              type:secret, alert:repo-owner
RESULT.CONTACT:           Example User <user@example.com>
RESULT.LOCATION.VERSION:  0964f2f08aed3b5c34d0de25609046149e5c1a28
RESULT.LOCATION.PATH:     generic/valid/secret-in-config
RESULT.LOCATION.RANGE:    L1C6-L1C102  
RESULT.NOTES.MESSAGE:     Merge of ...

ID:                       d2f80dc543822e4a319542914864dd7cd961cf8ef2e1d4bfe68b06871474fe5b
REQUEST.ID:               46a96aa1a87283d055b8d1e297b8fde379255d09058b008b4346384dea2bb470
RESULT.ID:                fde57c9bd7a0f2969dce4f7d526dfe550fa031e5711ed624c85d43e3e11a993d
RESULT.KIND:              GitCommit
RESULT.SECRET:            ...
RESULT.MATCH:             ...
RESULT.ENTROPY:           5.0 
RESULT.DATE:              ...
RESULT.RULE.ID:           cdcf75b34688c4e0d60fa7a6e76fb4ca38f979b3278c4bfed23ebb1366790e22
RESULT.RULE.DESCRIPTION:  Generic Secret
RESULT.TAGS:              type:secret, alert:repo-owner
RESULT.CONTACT:           Example User <user@example.com>
RESULT.LOCATION.VERSION:  0964f2f08aed3b5c34d0de25609046149e5c1a28
RESULT.LOCATION.PATH:     generic/valid/secret-in-config
RESULT.LOCATION.RANGE:    L1C6-L1C102 
RESULT.NOTES.MESSAGE:     Merge of ...

...
```

**Note:** A few of the fields have been merged together to be more concise. For example `RESULT.LOCATION` and `RESULT.CONTACT` are made up of separate fields in the json

thewizzy:
Resolves #84


pkg/response/response.go
```diff
@@ -19,65 +26,243 @@ const (
...
+}
+
+// flattenContact creates a single string with Contact information as "Name <Email>"
+func flattenContact(contact Contact) string {
+	return fmt.Sprintf("%s <%s>", contact.Name, contact.Email)
+}
+
+// sanitizeField takes a string and makes it safe for CSV by replacing newlines and escaping quotes
+// CSV is not a great format for rich/nested data, this makes the data more consistent for the format
+func sanitizeField(value string) string {
```
bplaxco, line 261: @thewizzy does the CSV writer not handle that automagically? Or is this more for the human one?

rgmz: CSV breaks with nested `'` or newlines. I was shocked at how little escaping it does.

cmd/cmd.go
```diff
@@ -194,6 +196,7 @@ func runListen(cmd *cobra.Command, args []string) {
 
 	// Prints the output of the scanner as they come
 	go leakScanner.Recv(func(response *response.Response) {
+		response.SetFormat(outputFormat)
```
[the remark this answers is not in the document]
thewizzy: This is how I implemented it at first. But the formatting is also so tightly connected to the response. 

We could create a struct for the Formatter inside `pkg/response/formatter.go', I think it is important to keep it with the response that its formatting. 
For example, the json, toml and yaml are annotated on the Response struct, its the human and csv that have some more related code.

examples/config.toml
```diff
@@ -4,6 +4,11 @@
 # All items in the config should have sane defaults and the
 # config is optional.
 #
+[formatter]
+
+# Valid values: "CSV", "HUMAN", "JSON", "TOML", "YAML"
+format = "JSON"
```
bplaxco, line 10: Do you have a preference between

What you have: (kinda matches how log levels are done)
```
"CSV", "HUMAN", "JSON", "TOML", "YAML"
```
vs

Lower case for easier typing (e.g. `leaktk-scanner scan --format=json ...`) and kinda feels more like how I've seen it in other tools.
```
"csv", "human", "json", "toml", "yaml"
```
vs

This which is kinda CamelCase like the "kind" values (e.g. GitRep, etc) since this is a type of thing. 
```
"CSV", "Human", "JSON", "TOML", "YAML"
```

I lean a little towards lower but the others also make sense in my mind so I'm not sold on a single format. I figured I'd ask but also what's here is perfectly valid too.

thewizzy: I prefer lower but was trying to match the log level case.

Regardless it will accept any mix of upper/lower case.

cmd/cmd.go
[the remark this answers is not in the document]
bplaxco: Okay, yeah, let's put it in `pkg/response/formatter.go` and since it'll only be for formatting response objects, we could change that from:

```go
fmt.Println(formatter.FormatResponse(response))
```
to 

```go
fmt.Println(formatter.Format(response))
```

examples/config.toml
bplaxco: Ack ack. I think we document it as lower and run with it :ok_hand: 

Since it's more of a CLI option than a normal setting, it just _feels_ like it should be lower by convention.

bplaxco:
```
format = "JSON"
```
and 
```
--format=json
```
feel right to me I think.

cmd/cmd.go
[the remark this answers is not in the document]
bplaxco: @thewizzy actually I commented on the wrong one. 

The line 100 should be `fmt.Println(formatter.Format(response))` and this one should be removed because the listen results should always be JSONL for batch processing.

bplaxco (changes requested):

pkg/response/formatter.go
```diff
@@ -0,0 +1,211 @@
...
+		return CSV, nil
+	default:
+		return JSON, fmt.Errorf("invalid output format option: format=%q", format)
+	}
+}
+
+// SetFormat sets the format using the OutputFormat value
+func (f *Formatter) SetFormat(format OutputFormat) {
+	f.format = format
+}
```
bplaxco, line 68:
```suggestion
```

This can be removed, correct?

It doesn't seem to be used anywhere and it doesn't seem like we're mutating the formatter after it's created.

pkg/response/formatter.go
```diff
@@ -0,0 +1,211 @@
...
+	flat := flattenedResponse(r)
+	var out []string
+	for _, response := range flat {
+		for i, entry := range response {
+			// Specifies width of 26 characters for labels
+			out = append(out, fmt.Sprintf("%-26s%s", headers[i], entry))
+		}
+		out = append(out, "\n")
+	}
+	return strings.Join(out, "\n")
```
bplaxco, line 111: Thoughts on using a strings builder?

```suggestion
	var out strings.Builder

	headers := flattenedResponseFields()
	flat := flattenedResponse(r)

	for _, response := range flat {
		for i, entry := range response {
			// Specifies width of 26 characters for labels
			fmt.Fprintf(&out, "%-26s: %s\n", headers[i], entry)
		}
		out.WriteRune('\n')
	}

	return out.String()
```

The output would look like this:

```
...
RESULT.ENTROPY            : 2.584963
RESULT.DATE               : 2024-08-14T18:24:03Z
RESULT.NOTES              : message: Add a yarn file
RESULT.RULE.ID            : QqS4RvI6Zmg
RESULT.RULE.DESCRIPTION   : Authorization Header
RESULT.RULE.TAGS          : "type:secret, group:leaktk-testing"
RESULT.CONTACT            : Braxton Plaxco <bplaxco@redhat.com>
RESULT.LOCATION.VERSION   : 0795d84ec8b38c6fe95cb2f0eba52525585c5486
RESULT.LOCATION.PATH      : yarn-4.2.2.cjs
RESULT.LOCATION.RANGE     : L696C4824-L696C4844

...
```

pkg/response/formatter.go
```diff
@@ -0,0 +1,211 @@
...
+
+	for _, result := range response.Results {
+		flattened = append(flattened, []string{
+			sanitizeField(response.ID),
+			sanitizeField(response.RequestID),
+			sanitizeField(result.ID),
+			sanitizeField(result.Kind),
+			sanitizeField(result.Secret),
+			sanitizeField(result.Match),
+			sanitizeField(result.Context),
```
bplaxco, line 170: I noticed that Context can get rather big some times. 

Thoughts on removing that and keeping match and secret? We can always tweak Match to be what we want to show the user in the rules.

pkg/response/formatter.go
```diff
@@ -0,0 +1,211 @@
...
+
+	var buf bytes.Buffer
+	writer := csv.NewWriter(&buf)
+
+	err := writer.Write(headers)
+	if err != nil {
+		logger.Error("could not write response: error=%q", err)
+	}
+
+	err = writer.WriteAll(flattenedResponse(r))
```
bplaxco, line 142: Thoughts on only doing the sanizieField on the CSV one?

```suggestion
        	for _, row := range flattenedResponse(r) {
		sanitizedRow := make([]string, len(row))
		for i, field := range row {
			sanitizedRow[i] = sanitizeField(field)
		}

		err = writer.Write(sanitizedRow)
		if err != nil {
			break
		}
	}

```

Rationale: 

For the human readable one, we'd want to show exactly what was matched for the sake of someone grepping for it.

Only issue for the human one I see is for multi-line values. But for gitleaks they handle it just by truncating things and letting it wrap:

```
Finding:     -----BEGIN PRIVATE KEY-----                                     
MIIEvQIBADANBgkqhkiG9w0BAQEFAASCBKcwggSjAgEAAoIBAQDW8OVrrcJ/rZxZ
/kHlgz5...                                                      
Secret:      -----BEGIN PRIVATE KEY-----                                     
MIIEvQIBADANBgkqhkiG9w0BAQEFAASCBKcwggSjAgEAAoIBAQDW8OVrrcJ/rZxZ
/kHlgz5...                                                      
RuleID:      private-key
Entropy:     6.028506
File:        pki/valid/private.key
Line:        2
Fingerprint: pki/valid/private.key:private-key:2
```

thewizzy: Now only sanatize on CSV. 

There is also a truncate option for truncating commonly long lines.

pkg/response/formatter.go
thewizzy: Kept all three but there is an option on human output to truncate output.

pkg/response/formatter.go
```diff
@@ -0,0 +1,232 @@
...
+// flattenedResponseFields provides a list containing the field labels for a flattened response
+func flattenedResponseFields() []string {
+	return []string{"ID", "REQUEST.ID", "RESULT.ID", "RESULT.KIND", "RESULT.SECRET", "RESULT.MATCH",
+		"RESULT.CONTEXT", "RESULT.ENTROPY", "RESULT.DATE", "RESULT.NOTES", "RESULT.RULE.ID", "RESULT.RULE.DESCRIPTION",
+		"RESULT.RULE.TAGS", "RESULT.CONTACT", "RESULT.LOCATION.VERSION", "RESULT.LOCATION.PATH",
+		"RESULT.LOCATION.RANGE"}
+}
+
+// flattenedResponseFields provides a list containing the field labels for a flattened response
+func truncatableResponseFields() []int {
```
thewizzy, line 159: Use this to create a small list of ints for which fields are to be truncated. This saves a lot of string compares

pkg/response/formatter.go
thewizzy: I have removed RESULT.CONTEXT from human format based on discussions

bplaxco (approved):
Added a few nits and suggestions. Feel free to make as many as you see fit and merge in. 

RE: My order changes

Running it a few times on nosy, real repos and looking at the output might help decide on what makes the best sense. I do think I like keeping the CSVs and human terse though just for the sake of skimming (human) and probably tuning (csv).

pkg/response/formatter.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, entry := range value {
+		if strings.ContainsAny(entry, ",\n\"") {
+			entry = strings.ReplaceAll(entry, "\"", "\"\"")
+			entry = strings.ReplaceAll(entry, "\n", " ")
+			entry = fmt.Sprintf("\"%s\"", entry)
+		}
+		output = append(output, entry)
+	}
+	return output
+}
```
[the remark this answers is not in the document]
thewizzy: Removed sanitize. It was mostly helping with terminal out, we have human output if people want that.

pkg/response/formatter.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		logger.Error("could not marshal response: error=%q", err)
+	}
+	return string(out)
+}
+
+func formatHuman(r *Response) string {
+	headers := flattenedResponseFields()
+
+	skipContext := 6 // Skip the RESULT.CONTEXT
```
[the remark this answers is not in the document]
thewizzy: Based on this comment and discussions

I dropped the following fields:
* RESULT.CONTEXT
* RESULT.NOTES

I return the headers and the responses from a single "flatten". Keeps them closer together if things do change. I didn't like the magic index either 💯 

I have reordered the output to match your recommendation.

bplaxco:
The output looks clean and so does the code :ok_hand: :ok_hand: 

Shipit! :shipit:

PR #95, opened 2025-02-12
Don't create config/cache dirs until needed

bplaxco:
This fixes an issue where you can get messages like:

```
could not create dir: path="/.cache/leaktk"
```

When you run leaktk-scanner as a subprocess, are passing in a specific workdir from the config and HOME isn't set. Before it would try to create ~/.cache/leaktk first, even if that's not where the cache dir was configured to be.

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


thewizzy (approved):
Good find!

PR #105, opened 2025-02-21
Support @/path for the --resource flag for ad-hoc scans

bplaxco:
**Issue**

JSONData scans from `leaktk-scanner scan` don't work when there's a lot of json data and I don't want to have to convert it to a request file.

**Fix**

If `--resource` starts with `@` assume it's referencing a file to load as the data for that field. This is similar to how curl does it.

Example usage of the change:

```
leaktk-scanner scan --kind JSONData --resource @data-file.json
```


thewizzy (approved):
LGTM. Nice and simple, tested with files on windows.

PR #106, opened 2025-02-21
Include gitleaks fingerprint

bplaxco:
Handy for if you need to write a .gitleaksignore for your repo. 

**Alt Idea**

(WIP: considering just creating a fingerprint field on the result object and passing it through but that depends on how stable it is in gitleaks and if we want to tie into that same format).

Review this after https://github.com/leaktk/scanner/pull/105 is merged.


thewizzy (approved):
If you have the use case for it. It only applies to git and goes into the notes.
It seems like it would be very consistent given these constraints.

bplaxco:
Sounds good. I'm going to go ahead and merge in and we can revisit any tweaks we want to make before 1.0. :ok_hand:

PR #112, opened 2025-02-26
Return a LeakExitCode when a leak is found on a single scan

thewizzy:
This is useful for automation, especially where we do not want to consume and redisplay the whole program out.

We could set this as a command line option too, like gitleaks does, if deemed necessary?


thewizzy:
Closing as ill lump it all in #113

PR #113, opened 2025-02-27
Changes to support scanning "staging" for precommit

thewizzy:
This PR brings in:
* Return exit code 42 on exit if there are leaks found
* Adds support for local git repos. Checks if the path exists and it contains a .git folder. (I think will need to check this against worktrees)
* Adds the gitrepo option `"staged": bool` this changes feed for the detector from git log to a git diff --staged. 

Outstanding / Feedback required:
* Need to pass redact through to the detector. The easiest way is to add this to the options for all types, and this would support redaction when running in listen. Another option, or alongside, is being able to specify it with --redact when running scanner. Adding to options seems clunky if you want to run single scans on the command line. Thoughts?
* Test and modify local git repo check to support worktrees.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -85,7 +85,7 @@ func logoutCommand() *cobra.Command {
 	}
 }
 
-func runScan(cmd *cobra.Command, args []string) {
+func runScan(cmd *cobra.Command, args []string) error {
 	request, err := scanCommandToRequest(cmd)
 
 	if err != nil {
@@ -95,16 +95,27 @@ func runScan(cmd *cobra.Command, args []string) {
 	var wg sync.WaitGroup
 	leakScanner := scanner.NewScanner(cfg)
 
+	// Create a channel for a result count
+	results := make(chan int)
 	// Prints the output of the scanner as they come
 	go leakScanner.Recv(func(response *response.Response) {
 		fmt.Println(response)
+		results <- len(response.Results)
+		close(results)
 		wg.Done()
 	})
 
 	wg.Add(1)
 	leakScanner.Send(request)
 
+	result := <-results
+
 	wg.Wait()
+	if result != 0 {
+		return errors.New("potential leak identified")
+	}
+
+	return nil
 }
 
 func scanCommandToRequest(cmd *cobra.Command) (*scanner.Request, error) {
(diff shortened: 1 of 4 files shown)


bplaxco:
Aside: I had started some local scan work here if any of that is useful:

https://github.com/leaktk/scanner/pull/74/files

Rationale behind a few changes:

- ClonePath -> Path where path just means the path of the resource on the file system since something like a directory or local repo isn't "cloned"
- A local flag with [this behavior](https://github.com/leaktk/scanner/pull/74/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R132) so that's possible for overriding some things in the future (i.e. scan local image or a remote image, etc)
- Gitleaks has some [functions for unstaged/staged items](https://github.com/leaktk/scanner/pull/74/files#diff-f8472fe0bafccc33f268c74d1bd2a697594914d4e1d796805b254249cbc78828R114).

Note:

- That branch is a bit stale so some of the diff (esp the logging bits) are outdated)
- That is a draft branch so I was still working through some of the ideas. Feel free to take or leave as much of that as you want

thewizzy:
> Aside: I had started some local scan work here if any of that is useful:
> 
> https://github.com/leaktk/scanner/pull/74/files
> 
> Rationale behind a few changes:
> 
>     * ClonePath -> Path where path just means the path of the resource on the file system since something like a directory or local repo isn't "cloned"
> 
>     * A local flag with [this behavior](https://github.com/leaktk/scanner/pull/74/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R132) so that's possible for overriding some things in the future (i.e. scan local image or a remote image, etc)
> 
>     * Gitleaks has some [functions for unstaged/staged items](https://github.com/leaktk/scanner/pull/74/files#diff-f8472fe0bafccc33f268c74d1bd2a697594914d4e1d796805b254249cbc78828R114).
> 
> 
> Note:
> 
>     * That branch is a bit stale so some of the diff (esp the logging bits) are outdated)
> 
>     * That is a draft branch so I was still working through some of the ideas. Feel free to take or leave as much of that as you want

I'll move to bringing this branch up to date. For some reason i had it in my mind that this branch was a new resource type. Ill rebase it on `update-gitleaks` branch. That should be ready to go in soon

PR #120, opened 2025-03-21
Support archives for `--kind Files`

bplaxco:
Support decoding archives for `--kind Files`, other types are currently out of scope but will be done in the future. 

Nice to have: Refactor Files/Archives/Text to use a generic FileSystem interface like archives does.


pkg/resource/files.go
```diff
@@ -84,35 +87,27 @@ func (r *Files) Since() string {
 
 // ReadFile provides a way to access values in the JSON data
 func (r *Files) ReadFile(path string) ([]byte, error) {
-	return os.ReadFile(filepath.Join(r.path, filepath.Clean(path)))
+	// Clean the path before passing it to the fs. This also turns "" -> "."
+	// which is important given we don't use "." to represent an empty path
+	file, err := r.fs.Open(filepath.Clean(path))
+	if err != nil {
+		return nil, err
+	}
```
thewizzy, line 95: Make sure the file is closed after reading the contents.
```suggestion
	}
	defer file.Close()
```

thewizzy:
The following line seems to be causing the test failure on windows. If you `FromSlash` it back then it will work.
https://github.com/mholt/archives/blob/main/fs.go#L800

It doesn't have issues reading the file content in the test. It is just the path separator change. 

It shouldn't matter, can update the test or convert it back as required. Windows will generally take both / and \ as separators

PR #128, opened 2025-04-04
Change the container to use alpine & add git

thewizzy:
Smallest container to get package manager and git - needed by leaktk-scanner.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,3 +1,5 @@
-FROM scratch
+FROM alpine/git
 COPY leaktk-scanner /usr/local/bin/leaktk-scanner
+RUN addgroup -S scanner && adduser -S scanner -G scanner
+USER scanner
 ENTRYPOINT ["/usr/local/bin/leaktk-scanner"]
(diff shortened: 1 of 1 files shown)


Containerfile
```diff
@@ -1,3 +1,5 @@
-FROM scratch
+FROM alpine/git
```
bplaxco, line 1:
```suggestion
FROM docker.io/alpine/git
```

Just so things like podman don't have to guess based on their registries.conf.

Containerfile
```diff
@@ -1,3 +1,5 @@
...
 COPY leaktk-scanner /usr/local/bin/leaktk-scanner
-ENTRYPOINT ["/usr/local/bin/leaktk-scanner"]
\ No newline at end of file
+RUN addgroup -S scanner && adduser -S scanner -G scanner
+USER scanner
```
bplaxco, line 4:
```suggestion
RUN \
  mkdir -p /var/lib/leaktk && \
  adduser -D -h /var/lib/leaktk/scanner -u 1001 -G root leaktk-scanner && \
  chmod -R ug=rwX,o-rwx /var/lib/leaktk/scanner

USER 1001
```

I think you'd want the user to be in the root group so it'd play nicer in OpenShift and such.

And somewhere in the back of my mind I feel like there was something about wanting to set a specific user ID.

Also throwing the home dir under /var/lib/leaktk as a place to standardize on for leaktk runtime app data.

Seem legit?

Issue #129, opened 2025-04-04
RFE: Set custom user agent


bplaxco:
Requests to the pattern server (and any other requests made by the scanner) should use a custom user agent:

```
leaktk-scanner/<version> (<os> <arch>)
```

thewizzy:
In #132 i took a stab at implementing this. I added an @<commit> after <version>.

bplaxco:
Fixed in #132

PR #132, opened 2025-04-10
Implement custom useragent

thewizzy:
This will override all http library user agents as long as they are using the http.DefaultTransport.

There is a way to set the container images library useragent (see pkg/resource/container_image.go). I am just trying to work out how to go from where we are to having it in the container_image file too. It is close to cmd because it uses version and commit from there. 

Example user-agent `leaktk-scanner/0.0.17@12312312 (linux amd64)`.

Implements #129

### DIFF
--- a/cmd/useragent.go
+++ b/cmd/useragent.go
@@ -1,0 +1,36 @@
+package cmd
+
+import (
+	"fmt"
+	"net/http"
+	"runtime"
+)
+
+// globalUserAgent is the useragent used by default by http
+var globalUserAgent = fmt.Sprintf("leaktk-scanner/%s (%s %s)", shortVersion(), runtime.GOOS, runtime.GOARCH)
+
+type userAgentTransport struct {
+	rt http.RoundTripper
+}
+
+func (uat *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req = req.Clone(req.Context())
+	req.Header.Set("User-Agent", globalUserAgent)
+	return uat.rt.RoundTrip(req)
+}
+
+func init() {
+	http.DefaultTransport = &userAgentTransport{
+		rt: http.DefaultTransport,
+	}
+}
+
+func shortVersion() string {
+	if len(Version) > 0 {
+		if len(Commit) > 0 {
+			return Version + "@" + Commit
+		}
+		return Version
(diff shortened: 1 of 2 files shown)


version/useragent.go
```diff
@@ -0,0 +1,40 @@
...
+	rt http.RoundTripper
+}
+
+func (uat *userAgentTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req = req.Clone(req.Context())
+	req.Header.Set("User-Agent", globalUserAgent)
+	return uat.rt.RoundTrip(req)
+}
+
+func init() {
```
bplaxco, line 22: hrm something about this only running when the package is imported feels off, and if anything ever used this as a library, it'd change the user agent of that code silently in the background.

What are your thoughts a `NewClient` function in `pkg/http/http.go` that returns a configured http client that should be used for things in the project making HTTP requests?

That'd give us the ability to tweak things as needed in different contexts and both protect us from other tools messing with the default client and other tools from us if our code is imported.

thewizzy: We could split useragent from version so that it is only imported by cmd, not but container_image as well. That would resolve it impacting others. 

Ill pull http.go out of pkg/scanner into pkg/http and go from there.

pkg/http/http.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"net/http"
+	"sync"
+
+	"github.com/leaktk/scanner/version"
+)
+
+// HTTPClient provides an interface for working with Go's http client or
+// swapping it out with other types for testing
+type HTTPClient interface {
```
thewizzy, line 12: @bplaxco is this used? I couldn't see any tests where it was swapped out.

bplaxco: Probably not if you're not seeing it anywhere :ok_hand:

bplaxco:
LGTM! And if that interface you saw above isn't used, feel free to rip it out and merge on it.

pkg/http/http.go
thewizzy: It was use but not practically used. Patterns used it but really only as a shim. It appears like it might have been a direction you were considering before you went with `httptest`.
I have stripped it out and added a user-agent check to the patterns tests.

PR #133, opened 2025-04-11
GCS Scanning Support

bplaxco:
Allow scanning GCS buckets for single objects (default) or multiple objects (if the query option is provided).


pkg/resource/gcs.go
```diff
@@ -0,0 +1,224 @@
...
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+
+	"github.com/leaktk/scanner/pkg/logger"
+	"github.com/leaktk/scanner/pkg/response"
+)
+
+var getClientMu sync.Mutex
+var storageClient *storage.Client
```
rgmz, line 20: Absolutely inconsequential, but you can group `var` declarations:
```suggestion
var (
    getClientMu sync.Mutex
    storageClient *storage.Client
)
```

pkg/resource/gcs.go
```diff
@@ -0,0 +1,224 @@
...
+
+// newStorageClient returns a usable storage client for resources. Since
+// storage.Client is safe across multiple threads, it returns the same
+// client each time
+func newStorageClient() (*storage.Client, error) {
```
rgmz, line 25: It seems like this would be a good use of [`sync.Once`](https://victoriametrics.com/blog/go-sync-once/).

pkg/resource/gcs.go
bplaxco: I always forget about that in go. Thanks for the note!

PR #141, opened 2025-05-30
Set custom exit code when leaks are found for scan

bplaxco:
Allow setting `--leak-exit-code` for CI use cases. (default 0)


abutcher (approved):
LGTM

PR #142, opened 2025-06-03
Rename leaktk-scanner -> leaktk

bplaxco:
This:

- Renames `leaktk-scanner` to `leaktk`
- Breaks apart the docs

After this is merged, I'll need to

- Update the container repo
- Update update this repo
- Update tools and docs in hack
- Update downstream tools


bplaxco:
Going to go ahead and pull this in so we can start building the other features on top of it :ok_hand:

PR #148, opened 2025-06-09
[WIP] Support archive scanning

bplaxco:
Scan files and traverse archives for paths ending in archive file names:

```
ID:                       YBq8q6H45kA
REQUEST.ID:               KJGJmNaQdEI
RESULT.ID:                9DoVUPlvX-U
RESULT.KIND:              General
RESULT.RULE.ID:           rnWF160pWNg
RESULT.RULE.DESCRIPTION:  Generic Secret
RESULT.CONTACT:            <>
RESULT.SECRET:            /ANvEGX1XY9BS/YNvE:asdf.YNvEGXEXY9BS}
RESULT.MATCH:             Token: /ANvEGX1XY9BS/YNvE:asdf.YNvEGXEXY9BS}
RESULT.ENTROPY:           4.033514
RESULT.DATE:              
RESULT.LOCATION.VERSION:  
RESULT.LOCATION.PATH:    fake-leaks.tar.gz!fake-leaks/yaml-test.yml
RESULT.LOCATION.RANGE:    L8C8-L8C51
RESULT.RULE.TAGS:         type:secret, alert:repo-owner
```

**To Do**

- [ ] Get existing tests passing using gitleaks sources for scanning (**WIP**)
- [ ] Write new tests to cover archive testing
- [ ] Another round of refactoring


pkg/config/config.go
```diff
@@ -178,6 +179,7 @@ func DefaultConfig() *Config {
 			MaxScanDepth:        0,
 			ScanWorkers:         1,
 			Workdir:             filepath.Join(xdg.CacheHome, "leaktk", "scanner"),
+			MaxArchiveDepth:     8,
```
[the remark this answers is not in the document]
bplaxco: @rgmz sadly I don't have any hard data. Based on what I've anecdotally seen in the wild, I think 1 is the most common and then 2 happens from time to time. 8 was my: "this shouldn't blow up memory or processing time but it also shouldn't miss stuff" guess.

It'd be interesting to pass that archive depth to the Fragment and include that in the tags for reporting though :thinking: 

@zricethezav would you want archive tagging similar to how decoding adds tags upstreamed?

thewizzy:
The way container_image scanning worked before this up was that it would select 1 platform/arch to download and scan. The current setup scans all containers in the manifest and all layers.

Do we want to revert back to selecting a single platform/arch? or scan every one in the main manifest?

Otherwise this tests pretty close. I have yet to compare v0.1.0 to this programmatically

thewizzy:
> Woot! One thing that I expect will be different is the `!` separator for things like this JSON referenced a URL and that was fetched and any other places where you "cross a boundary" between one source to another or nested source. I haven't worked though all the cases yet to see if they're consistent, but that's kind of the way I've been thinking about it so far.
> 
When a url is referenced in JSON the path is the name of the json field that contains the url. It does not share the actual URL. Was this expected to look something like `<field name>!<url>` ?

bplaxco:
> When a url is referenced in JSON the path is the name of the json field that contains the url. It does not share the actual URL. Was this expected to look something like `<field name>!<url>` ?

So if you had:

```json
{
   "data": {"creds": "https://example.com/secret.json"}
}
```

where `https://example.com/secret.json` returned:

```json
{
  "secret": "super secret password!"
}
```

The path would be:

```
data/creds!secret
```

But if it was something like this where the response was just flat text instead of JSON:

```json
{
   "data": {"creds": "https://example.com/secret.txt"}
}
```

where `https://example.com/secret.txt` returned:

```
secret="super secret password!"
```

You'd end up with just:

```
data/creds
```
because there's no sub-path to consider (kind of like how gitleaks does for a compressed file vs an archive file that has inner paths).

Thoughs/see any issues with that approach?

thewizzy:
> Thoughs/see any issues with that approach?

That works. It is how it currently is. I added a test to json_source_test.go to cover this. It only works in cases where the server responses with the content-type application/json - so I am not sure how often we will hit this but its ok.

pkg/scanner/scanner.go
```diff
@@ -1,188 +1,506 @@
...
-				Priority: msg.Priority,
-				Value: &response.Response{
-					ID:        id.ID(),
-					Results:   make([]*response.Result, 0),
-					Logs:      reqResource.Logs(),
-					RequestID: request.ID,
-				},
+		logger.Info("starting scan: id=%q", request.ID)
+
+		// TODO: make sure the timeout is handled properly everywhere
```
bplaxco, line 106: @thewizzy can this comment be removed or are there still things to check?

pkg/scanner/scanner.go
```diff
@@ -1,188 +1,506 @@
...
+			result.Notes["image"] = request.Resource + "@" + manifest
+		} else {
+			result.Notes["image"] = request.Resource
+		}
+		// TODO: handle the different kinds like image vs manifest
+		// and sub images and such here where the path needs to be re-mapped
+		// and you can tell the kind based on the path prefix (e.g. layers, manifest, etc)
+		//
+		// Also fix this not returning results:
+		//  ./leaktk scan --kind ContainerImage "docker://quay.io/leaktk/fake-leaks:v1.0.1"
```
bplaxco, line 369:
```suggestion
```

This was resolved :ok_hand:

pkg/scanner/scanner.go
```diff
@@ -1,188 +1,506 @@
...
+			result.Kind = proto.ContainerMetdataResultKind
+		}
+		if manifest != "" {
+			result.Notes["image"] = request.Resource + "@" + manifest
+		} else {
+			result.Notes["image"] = request.Resource
+		}
+		// TODO: handle the different kinds like image vs manifest
+		// and sub images and such here where the path needs to be re-mapped
+		// and you can tell the kind based on the path prefix (e.g. layers, manifest, etc)
```
bplaxco, line 366: Anything else to do here or can this be removed?

bplaxco (commented):
LGTM! Added a few questions about TODO comments and @abutcher will take a look as well ^_^

PR #154, opened 2025-08-04
Add leaktk scan --gitleaks-config to provide a custom gitleaks config

abutcher:
Add `--gitleaks-config` flag for overriding the Gitleaks config `leaktk scan --gitleaks-config <path-to-config>`.

```
LEAKTK_LOGGER_LEVEL=DEBUG ./leaktk scan --gitleaks-config /tmp/wahoo --kind Files /tmp/hehe
[DEBUG] using provided gitleaks config: path=/tmp/wahoo
[DEBUG] disabling pattern autofetch with custom gitleaks config
[DEBUG] disabling pattern expiredafter/refreshafter with custom gitleaks config
[INFO] queueing scan: id="Mc6lCK73nck"
[INFO] starting scan: id="Mc6lCK73nck"
[DEBUG] no additional config
[DEBUG] gitleaks: scanning path
[INFO] queueing response: id="Mc6lCK73nck"
...
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -226,7 +226,7 @@ func scanCommand() *cobra.Command {
 	flags.StringP("kind", "k", "GitRepo", "the kind of resource to scan")
 	flags.StringP("options", "o", "{}", "additional request options formatted as JSON")
 	flags.Int("leak-exit-code", 0, "the exit code when leaks are found (default 0)")
-	flags.StringP("gitleaks-config", "g", "", "the path to a custom gitleaks config")
+	flags.StringP("gitleaks-config", "", "", "the path to a custom gitleaks config")
 
 	return scanCommand
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -85,6 +85,17 @@ func runScan(cmd *cobra.Command, args []string) {
 		logger.Fatal("invalid leak-exit-code: %v", err)
 	}
 
+	gitleaksConfig, err := cmd.Flags().GetString("gitleaks-config")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/cmd.go
```diff
@@ -85,6 +85,17 @@ func runScan(cmd *cobra.Command, args []string) {
...
 
+	gitleaksConfig, err := cmd.Flags().GetString("gitleaks-config")
+	if err != nil {
+		logger.Fatal("invalid gitleaks-config: error=%q", err.Error())
+	}
+
+	// Providing a gitleaks-config via command line arguments takes
+	// precedence over gitleaks config set in the leaktk config file
+	if len(gitleaksConfig) != 0 {
+		cfg.Scanner.Patterns.Gitleaks.ConfigPath = gitleaksConfig
```
bplaxco, line 96: With a custom config passed in, we probably want to set auto fetch to false and disable the expiration check 👌

Just to keep it from overwriting the passed in file or freaking out of it has an old timestamp.

cmd/cmd.go
```diff
@@ -85,6 +85,22 @@ func runScan(cmd *cobra.Command, args []string) {
...
+		logger.Debug("using provided gitleaks config: path=%s", gitleaksConfig)
```
abutcher, line 97: Should this log be near here instead? https://github.com/leaktk/leaktk/blob/cc536e70a414d5c72dffc52be0ff78b840612ac5/pkg/scanner/patterns.go#L139

abutcher: Tentatively moved the log to there.

bplaxco (approved):
LGTM!

PR #159, opened 2025-08-13
Log fatal error when error on response for single scan

bplaxco:
When single scans were failing due to a misconfiguration in some tests, the exit code was 0 when it shouldn't have been.

This also implements the `error` interface for `proto.Error`


abutcher (approved):
LGTM

PR #177, opened 2025-09-23
Fix cases that can cause long scans

bplaxco:
Items:

- [x] respect max scan depth
- [x] since takes precedence over depth for clone but both are applied for scan
- [x] address scan timeout bug where timeout is ignored after the clone
- [x] set and kill entire process groups for git when timeouts are met to avoid child processes from hanging the goroutine.

### DIFF
--- a/pkg/scanner/git_command_unix.go
+++ b/pkg/scanner/git_command_unix.go
@@ -1,0 +1,22 @@
+//go:build !windows
+
+package scanner
+
+import (
+	"context"
+	"os/exec"
+	"syscall"
+)
+
+// gitCommand sets extra things on the command like the pgid
+// and cancel function to ensure the command doesn't hang
+// when in weird states
+func gitCommand(ctx context.Context, args ...string) *exec.Cmd {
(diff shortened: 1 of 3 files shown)


thewizzy:
Were these git commands failing? Should we have a retry facility? Re
Add it to the queue rather than just dropping?

bplaxco:
> Were these git commands failing?

@thewizzy, based on what I saw in the stderr for the git commands in the from that core dump and only seeing `git-remote-http` and not `git` in the ps output you sent, it looked like that was the case. Also it seemed like the `git-remote-http` processes were sleeping and note being reaped like they should.

> Should we have a retry facility? Re Add it to the queue rather than just dropping?

The above fix, should return a "signal killed" error, which should be returned in the response by the scanner. 

I think for the moment I like the idea of letting the caller handle that unless that ends up being a problem. Would you agree?

thewizzy:
@bplaxco reporting failures and letting the caller handle retry if they wish makes sense.

PR #208, opened 2026-02-02
`leaktk hook git.*` commands

bplaxco:
# Overview

This implements some of the features in https://github.com/leaktk/leaktk/issues/238:

- The scaffolding for future hooks and installers
- Git pre-commit and pre-receive hooks and installer
- Docs referenced by those hooks

# TODO

- [x] Finish the installer
- [x] Test git.pre-commit
- [x] Test git.pre-receive
- [x] Ensure separate leaktk processes can't write the patterns file at the same time.
- [x] Make sure all the doc link refs use the commit of the built version (or fall back on HEAD)
- [x] Have pre-commit properly include the .gitleaks.toml from the repo
- [x] Make sure the tests pass
- [x] Write the docs referenced by links
- [x] Do final spot check before marking it as ready for review

## Test Cases

Integration test cases can be added to `./test/integration` now.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -1,0 +1,140 @@
+package git
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/leaktk/leaktk/pkg/config"
+	"github.com/leaktk/leaktk/pkg/logger"
+)
+
+// GitRepoInfo is a collection of facts about a repo being scanned.
+// See `man 7 gitglossary` for more information about the terms.
+type RepoInfo struct {
+	// Whether or not the repo is a bare repo
+	IsBare bool
+	// The path to the actual GIT_DIR folder
+	GitDir string
+	// The working tree for the repo (a temp one is created for bare repos)
+	WorkingTreePath string
+}
+
+func GetRepoInfo(ctx context.Context, path string) (RepoInfo, error) {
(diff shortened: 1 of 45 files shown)


pkg/docs/docs.go
```diff
@@ -0,0 +1,49 @@
...
+	var b strings.Builder
+	b.WriteString(BaseDocsURL)
+	// Make sure there's a trailing slash
+	if BaseDocsURL[len(BaseDocsURL)-1] != '/' {
+		b.WriteByte('/')
+	}
+	b.WriteString(string(topic))
+	b.WriteString(DocsExt)
+	docURL := b.String()
+	if strings.Contains(docURL, "${GIT_REF}") {
```
bplaxco: Note to self: Probably don't want to pin to a specific ref actually because it'll make fixing docs for users less impactful.

internal/git/git.go
```diff
@@ -0,0 +1,140 @@
...
+		"--is-bare-repository",
+	) // #nosec G204
+
+	logger.Debug("executing: %s", cmd)
+	rawInfo, err := cmd.Output()
+	if err != nil {
+		return info, err
+	}
+
+	fields := bytes.Fields(rawInfo)
```
thewizzy: This in its current form fails when there is a space in the path.
```suggestion
	fields := bytes.Split(bytes.TrimSpace(rawInfo), []byte("\n"))
```

bplaxco: Good catch.

Issue #99, opened 2025-02-17
RFE: Support `.well-known` configuration for pattern server


bplaxco:
# Issue

`leaktk-scanner login` doesn't provide much information about where to get the auth token from and it might be different per each pattern server URL.

It does show

```
[INFO] logging in: pattern_server="<pattern-server-url>"
```

but that doesn't say whether or not the server even requires login or not.

# Idea

Support an optional `.well-known/leaktk-configuration` file that contains JSON like:

```json
{
  "patterns": {
      "token_endpoint": "<url-to-retrieve-token>",
      "grant_types_supported": [
        "manual",
      ]
  }
}
```

Where manual implies that you need to copy/paste the token from the screen to the prompt or something like that.

# Alt Idea (maybe preferred):

Support `<pattern-server-url>/.well-known/oidc-configuration` and just support OIDC directly in the pattern server.

Then `leaktk-scanner login` would do:

1. Check if server has oidc config
2. If so, do an OIDC flow for the user
3. Else, prompt with `Enter authorization header value` or something similar to set the raw auth header

Eventually supporting a `.well-known/leaktk` folder for things like Gitleaks config and such still might be a thing for URL scans for sites in a the future though. So we might could do both or a mix of the two. I'm up to hash it out with folks.

thewizzy:
I don't think the original idea is good. The `grant_type_supported` of `manual` is not in the specification - we want to keep within spec for something like this.

Ideally it would be good to implement one of the authentication flows, `authorization_code` might be the best. For those setting up a local system it will be fairly seamless, it would need to open the browser for the authentication but then the rest could be automated. 

Overall, most of this work will likely be on the pattern server side. We can use `golang.org/x/oauth2` here.

bplaxco:
>  The grant_type_supported of manual is not in the specification

Well that was supposed to be a custom file for leaktk not part of OIDC. The alt idea was just to just straight up support OIDC. 

`.well-known/leaktk-configuration` is meant to be what ever the tooling needs. I just borrowed a lot from OIDC. And the more I borrowed the more I considered just straight up supporting OIDC. But it still might be prefered to have a:

.well-known/leaktk-configuration

So pattern servers can provide context to tools about how to interact with it.

thewizzy:
Sorry. I had misunderstood the intentions here. 
This should work and be fairly simple to implement, more so than the alternative - though this could be a good stretch.

bplaxco:
[...]

PR #216, opened 2026-03-04
LeakTK-Analyst

bplaxco:
This was work pulled in from https://github.com/leaktk/leaktk/pull/181 

Most of this original work is by @alayne222 

# Context

Spec'ing out all of the details (that I can remember) from our conversations + answering some different questions about the models and analyst rules

# Patterns/Policy bundle discovery

- We can call this a "bundle" generically since it might hold patterns, policy etc as things evolve over time.
- We can use the [`.well-known/leaktk` idea](https://github.com/leaktk/leaktk/issues/99) for:
  - Specifying where to go get public keys for checking signatures
  - Providing links to specific bundles
  - Providing links to where patterns can still be downloaded

# Bundle vs Patterns

I think it'd make sense to publish our rego policies + patterns in a generic `bundle.tar.gz` format together. We can continue to support `/patterns/<provider>/<version>` for a while until we've had time to migrate everything over to the new format and then mark it as deprecated.

Things like the pattern server can also be updated to generate parts of bundles on the fly for when certain patterns require special group access.

## Logic for pulling bundle and patterns

1. Check if the server has a `.well-known/leaktk` URL
2. If not, skip to step 9.
3. Check if bundle URL is defined
4. If it is defined, pull the bundle
5. Check if the bundle has the patterns for the desired provider/version
6. If it does, stop
7. If it does not, check the well-known response for a pattern server URL
8. If it has one, use that as base URL for looking up the patterns and stop
9. If it doesn't assume the same base url that `.well-known` was appeneded to is the pattern server URL and try it

# Bundle layout

For building the bundle we can create a `src` directory with this layout:

```
src
├── betterleaks
│   └── v1_6_0
│       └── data.json
├── gitleaks
│   └── v8_27_0
│       └── data.json
└── leaktk
    └── v0
        └── analyst
            ├── models
            │   └── data.json
            └── policy.rego
```

And run `opa bundle -b src` to create a `bundle.tar.gz` that gets uploaded to the releases.

We can use OPA's signing features to include a signature in the bundle as well.

The layout here is `<provider>/<version>/<content>`. The existing gitleaks/betterleaks rules can be converted to json. This allows us to query `data.gitleaks.v8_27_0` for the rules allowlists etc.

# AI Features

Those have been moved to https://github.com/leaktk/leaktk/pull/221

# Analyst Details

The analyst feature mainly aims to take existing results and do any of the following with them:

- Filter results (i.e. remove findings based on a policy and leaktk config)
- Modify/enrich them (i.e. add analysis metadata, validation status through API calls, etc)
- Reference them (i.e. use them as context while doing any of the other actions to other results)

Getting started we're exploring OPA Rego policies for a lot of the logic to do the above + different ML models (initially a logistic regression) to enrich initially and then likely filter after it's a bit more proven in the wild.

# Design

I'm thinking of all of the bits of what we have in flight here as being pretty similar to existing gitleaks rules in that they're mainly here to dynamically configure what the scanner is looking for. So the models data and the rego policy will be provided by the same pattern server.

## Config Details

Update: Since we can use a .well-known url to lookup the details, we should be able to simplify the Patterns struct to just this again without a LeakTK specific config bit on it. If the concept of a bundle ends up working well over time we can tweak the names of these structs later.

```go
	Patterns struct {
		Autofetch    bool          `toml:"autofetch"`
		ExpiredAfter int           `toml:"expired_after"`
		Gitleaks     ProviderPatternsRef      `toml:"gitleaks"`
		RefreshAfter int           `toml:"refresh_after"`
		Server       PatternServer `toml:"server"`
	}
```

## Storage Details

Bundles should be stored at `~/.cache/leaktk/bundle.tar.gz`.

To maintain existing patterns compatibility and logic, if patterns are extracted from the bundle, they should be extracted, formatted as TOML, and stored at `~/.cache/leaktk/scanner/patterns/${provider}/${format_version}` like they are today. This way falling back on a pattern server still works as expected.

### Models

The `leaktk/v0/analyst/models/data.json` file can be formatted like this, where we define model refs for injecting into the analyst rego policy.

```javascript
{
   "models": [  // This will be used by https://github.com/leaktk/leaktk/pull/221
       {
            "kind": "LogisticRegression",
            // etc...
      }
   ],
}
```

### Analyst Policy

The analyst policy should take a dict like this for it's input field:

```javascript
{
  "response": { /* A LeakTK response object containing results and other metadata */ }
}
```

And to analyze the response, you should query:

```
data.leaktk.v0.analyst.analyzed_response
```

This should allow the policy to do anything it wants with the response and the results. Any result level analysis should be stored at `result.analysis`.

# Note

This is an updated version of the description based on a more recent discussion. The main thing is switching to the bundle format and making it the source for pulling patterns too. I think that having a generic "leaktk bundle" concept that has data in it that LeakTK can use to do its work will be flexible enough to grow with our use cases. For users wanting to just use custom gitleaks/betterleaks rules locally, we can keep the `--(gitleaks|betterleaks)-config` flags and maybe even make a `leaktk compile` command to help folks build their own bundles. Also if we switch away from Rego, the format should still be generic enough to put different content types in there.

# TODO

- [ ] *Important:* Do threat modeling given the flexibility of rego
- [ ] *Important:* Determine the signing process for the patterns and how that will be distributed
- [ ] Still more refactoring in `pkg/patterns` to dedup some common logic
- [ ] Make sure the wording is "patterns" and not "config" for what's fetched from the pattern server
- [ ] Work on pulling back in logistic regression work and exposing it through `leaktk.ai.RunModel(model, finding)` in the Rego policy
- [ ] Make sure things are tested well
- [ ] Move some analyzers from hack to the patterns server
- [ ] Test and final refactors


pkg/analyst/analyst.go
```diff
@@ -0,0 +1,67 @@
...
+// NewAnalyst initializes the Analyst with patterns.
+func NewAnalyst(p *patterns.Patterns) (*Analyst, error) {
+	return &Analyst{
+		patterns: p,
+	}, nil
+}
+
+func (a *Analyst) Analyze(ctx context.Context, response *proto.Response) (*proto.Response, error) {
+	// Get the latest LeakTK patterns
+	leaktkPatterns, err := a.patterns.LeakTK(ctx)
```
bplaxco, line 28: We'll need to find a way to provide ai.RunModel(modelName, finding) to this so it can link it in when it's compiling the pattern. One thing I had in mind ways something like this maybe:

```go
leaktkPatterns, err := a.patterns.LeakTK(ctx, patterns.WithFunction("leaktk.analyst.ai.RunModel", func(modelName string, finding proto.Result) {

}))
```

so that you can do the go pattern of these WithSomething functions as a varargs type argument to the function e.g.:

```go
runModel := patterns.WithFunction("leaktk.analyst.ai.RunModel", func(...) {

})

fooBar := patterns.WithFunction("leaktk.analyst.util.FooBar", func(...) {

})

...

leaktkPatterns, err := a.patterns.LeakTK(ctx, runModel, fooBar, ...)
```

still need to hash that out though and think through it more.

pkg/patterns/leaktk.go
```diff
@@ -0,0 +1,113 @@
...
+// LeakTK returns the LeakTKPatterns object, handling fetch/caching/update.
+func (c *Patterns) LeakTK(ctx context.Context) (*LeakTKPatterns, error) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	cfg := c.patternsConfig
+	localPath := cfg.LeakTK.LocalPath
+	modTimeExceeds := c.configModTimeExceeds(localPath, cfg.RefreshAfter)
+
+	if cfg.Autofetch && modTimeExceeds {
```
bplaxco: This and the similar code in gitleaks.go could use some further refactoring, most of this is the exact same still. Really the only thing that changes between the two is which URL is used to fetch things, which hash it's checked against, and which file/var are used.

pkg/ai/analyst.go
bplaxco, line 1: We can move `ai` under `pkg/ai` and then this basically just becomes a package that, for now, exports some structs and one function `RunModel` or something similar and then we can reference that inside patterns to expose `RunModel` to the rego there so it can pass it in and the analyst can leverage AI as it needs it through the rego policies. 

Would probably be good to expose a `HasModel` too so we can make things like running certain models conditional in the policy based on if the model exists or not.

bplaxco: Also if rego prefers snake case for it's functions we can map it to run_model and has_model, etc.

bplaxco: And we'd expose it under `leaktk.ai` instead of `leaktk.analyst.ai`

PR #232, opened 2026-04-15
Update macOS notarization configuration in goreleaser

thewizzy:
The notarization configuration is incorrect. Since id was added to builds it doesn't seem to find them. This /should/ bridge the gap.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -17,7 +17,10 @@ builds:
 
 notarize:
   macos:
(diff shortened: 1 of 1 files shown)


bplaxco:
Huh, the [docs](https://goreleaser.com/customization/sign/notarize/) say that the `ids` field should default to the project name. But lets give it a go!

bplaxco (approved):
LGTM!

PR #239, opened 2026-04-23
Refactor git hook installer

Eteriss:
## Summary

- Replace `--dir`/`--recursive` with `--path` (always recursive), rename template dir flags to match PR spec (`--user-template-dir`, `--system-template-dir`), add `-f` shorthand for `--force`
- Rewrite hook script template: adds `command -v` guard and error doc link instead of unconditionally exec-ing leaktk
- Replace `gitFindRepos` with `findGitDirs` using `filepath.WalkDir` (handles bare repos, always recursive, simpler force/skip logic in caller)
- Add `gitUserTemplateDir` to create and configure `init.templateDir` if not already set
- Add `docs/errors/command_not_found.md`
- Update tests to cover new flag API, template dir creation, and mtime-based force/skip behaviour

## Test plan

- [ ] `go test ./pkg/installer/...` passes
- [ ] `leaktk install hook git.pre-commit` installs hook in current repo
- [ ] `leaktk install hook git.pre-commit --path ~/somedir` installs in all repos under that path
- [ ] `leaktk install hook git.pre-commit --user-template-dir` creates/configures `init.templateDir` and installs hook there
- [ ] `--force` / `-f` overwrites existing hooks; without it existing hooks are skipped


bplaxco:
FYI: I pulled in a few changes from the original hook branch and made a few slight tweeks in https://github.com/leaktk/leaktk/pull/239/commits/cb4914adaad0650748e197cc19aba2b84bfe3e28.

That messed with the formatting a bit, but I'll run a `make format` after the merge to handle those :+1:

bplaxco:
There might be a few tweaks I make here and there but thanks for the help!

PR #242, opened 2026-05-07
Improve/fix the mac signing.

thewizzy:
Adding entitlements should tell it we are doing network and calling other binaries. 
Adding the certs to the chain as that should help too

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,12 @@ jobs:
         
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # pin@v4.0.0
+
+      - name: Import Code-Signing Certificates
+        uses: Apple-Actions/import-codesign-certs@v3
+          with:
+            p12-file-base64: ${{ secrets.MACOS_SIGN_P12 }}
+            p12-password: ${{ secrets.MACOS_SIGN_PASSWORD }}
         
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@ec59f474b9834571250b370d4735c50f8e2d1e29 # pin@v7.0.0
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -48,6 +48,12 @@ jobs:
         
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # pin@v4.0.0
+
+      - name: Import Code-Signing Certificates
+        uses: Apple-Actions/import-codesign-certs@v3
```
bplaxco, line 53:
```suggestion
        uses: Apple-Actions/import-codesign-certs@63fff01cd422d4b7b855d40ca1e9d34d2de9427d # pin@v3
```

entitlements.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
+    <key>com.apple.security.cs.allow-jit</key>
```
bplaxco, line 9: @thewizzy can you walk me through the specific cases for these?

bplaxco (changes requested):
The main change is pinning the dep, but also wanted to better understand which case each entitlement goes to. It seems like `com.apple.security.cs.allow-unsigned-executable-memory` might make `com.apple.security.cs.allow-jit` unnecessary but still reading through the docs. 

I take it those are for some of the wasm features in the regex engine used by betterleaks?

thewizzy:
Cancelling PR. Will look at using native code signing with goreleaser instead

PR #243, opened 2026-05-08
Avoid Mac notarization

bplaxco:
~There seems to be an issue with quill's signatures that we need to work through and this is a workaround for now until we can figure it out.~

A lot of the problems go away if we don't notarize it and just strip the quarantine flag. Home brew will be automated and I updated the docs for the users.


abutcher (approved):
LGTM

PR #244, opened 2026-05-11
Fix issues with archives in container scans

bplaxco:
- Upgrade mholt/archives to handle misidentifying tar archives
- Only scan regular files in containers
- Add extra logging
- Don't abort whole scans on archive errors


abutcher (approved):
LGTM

PR #247, opened 2026-05-12
Handle OCI image-spec v1 index manifests

bplaxco:
Handle  "application/vnd.oci.image.index.v1+json" manifest types.


abutcher (approved):
LGTM

PR #249, opened 2026-05-13
Fix rar panic and handle panics

bplaxco:
- **Replace rardecode with patched version**
- **Add panic handling in the scanner**


abutcher (approved):
LGTM

PR #251, opened 2026-05-14
Provide --stdout option for hook install && upgrade some deps

bplaxco:
To make it easier to automate installing hooks for cases where they're not installed in regular git dirs (e.g. GitLab's custom hooks dir). This provides a `--stdout` option in the installer so that it can handle generating the hook script content still but the automation doing the deployment can still decide where it goes.

**Example:**

Running this:
```sh
leaktk install hook git.pre-commit --stdout
```

Prints this to stdout:
```sh
#!/bin/sh
# TemplateID: f2998aee-4684-4c46-b724-7c8e37e6020c
# CreatedBy: leaktk-0.3.2+2-commits
# CreatedOn: 2026-05-14T16:33:35Z
if command -v leaktk > /dev/null 2>&1
then
    exec leaktk hook git.pre-commit
else
    echo 'leaktk command not found' >&2
    echo 'See: https://github.com/leaktk/leaktk/blob/HEAD/docs/error_command_not_found.md' >&2
    exit 1
fi
```


abutcher (approved):
LGTM

PR #252, opened 2026-05-18
Update src install

bplaxco:
There was an issue with the source install due to a replace statement in go.mod.

This change:

- Removes the replace statement
- Adds a doc for build dependencies 
- Updates the install commands in the Makefile to be macOS compatible
- Updates the grep command in hack/version to be macOS compatible
- Favors running `make` instead of `go install` to avoid similar issues in the future if we need to do a replace or automate some other build details.


abutcher (approved):
LGTM

PR #255, opened 2026-05-25
Implement GitHub source for monitor command

Eteriss:
## Summary
- Port GitHub live event stream from PwnedAlert to Go
- Add BearerToken auth support under `pkg/sources/auths/`
- Source IDs are now passed as CLI args instead of config

## Notes
- `--format=raw` is defined but not yet implemented


pkg/sources/github.go
```diff
@@ -1,10 +1,104 @@
...
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			logger.Error("could not read response body: %v", err)
+			continue
+		}
+		resp.Body.Close()
+
+		var events []gitHubEvent
+		if json.Unmarshal(body, &events) != nil {
+			logger.Error("could not parse github events: %v", err)
```
sharman-rh, line 92: This isn't the right error being logged, `err` was already logged on line 85

pkg/sources/auths/auths.go
```diff
@@ -0,0 +1,18 @@
...
+type BearerToken struct {
+	Kind  string
+	Token string
+}
+
+func NewAuth(Auth map[string]string) (BearerToken, error) {
+	if Auth["kind"] != "BearerToken" {
+		return BearerToken{}, fmt.Errorf("unrecognised auth kind: kind=%q", Auth["kind"])
+	}
+
```
sharman-rh, line 16: This only checks if the Auth kind is a Bearer token, not if the Bearer token actually has any value to it

pkg/sources/github.go
```diff
@@ -1,10 +1,104 @@
...
+		if newPoll != "" {
+			if seconds, err := strconv.Atoi(newPoll); err == nil {
+				poll = time.Duration(seconds) * time.Second
+			}
+		}
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			logger.Error("could not read response body: %v", err)
+			continue
```
sharman-rh, line 86: Needs `resp.Body.Close()` before `continue`

PR #267, opened 2026-06-23
Adding leaktk scan kinds in usage with examples

ravitri:
Adding `leaktk scan` kinds that are supported with some examples for easier/direct reference.


bplaxco:
Thanks for the contribution! Looks good to me.

PR #273, opened 2026-07-01
Redact command

alayne222:
Adding a `leaktk redact` feature. For now it can only handle stdio with the command `leaktk redact --kind Stdio`.


pkg/scanner/redact.go
```diff
@@ -0,0 +1,96 @@
...
+	detector.FollowSymlinks = false
+	detector.IgnoreGitleaksAllow = false
+	detector.MaxArchiveDepth = s.maxArchiveDepth
+	detector.MaxDecodeDepth = s.maxDecodeDepth
+	detector.MaxTargetMegaBytes = 0
+	detector.NoColor = true
+	detector.Redact = 0
+	detector.Verbose = false
+
+	buf := make([]byte, 2048)
```
bplaxco: Probably want to set that to something higher like `64 * 1024` just to reduce the chance of splitting large multiline secrets.

cmd/cmd.go
```diff
@@ -341,6 +341,39 @@ func versionCommand() *cobra.Command {
...
+	redactionMark := mustGetString(flags, "redaction-mark")
+	redactionWord := mustGetString(flags, "redaction-word")
+
+	leaktkScanner := scanner.NewScanner(cfg)
+
+	err := leaktkScanner.RedactStream(cmd.Context(), os.Stdin, os.Stdout, kind, redactionMark, redactionWord)
+	if err != nil {
+		logger.Fatal("redact stream execution failed: %v", err)
+	}
+}
```
bplaxco, line 487: I think I'd like to keep the redact feature separated from the scanner code since it'll grow over time and will also have a feature in the future to take existing findings and apply them to some resource.

What are your thoughts on something like this with the redaction code being in `pkg/redactor`:


**Note:** Regarding response.Resource: It's a new field on the Response struct and we'll want to make sure it has `json:"-"` on it so that it's not displayed for normal leaktk scans. It's kind of there to help pass along that context.


```suggestion
func yieldChunks(ctx, r io.Reader, yield func(chunk []byte, err error) error) error {
    buf := make([]byte, 64*1024)

    for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
			n, err := r.Read(buf)
            if err := yield(buf[:n], err); err != nil {
                return err
            }
    }
}

func runRedact(cmd *cobra.Command, args []string) {
    flags := cmd.Flags()
    kind := mustGetString(flags, "kind")

    // Update config from flags
    if redactionMark := mustGetString(flags, "redaction-mark"); redactionMark != "*" {
        cfg.Redactor.RedactionMark = redactionMark
    } 
    if redactionWord := mustGetString(flags, "redaction-word"); redactionWord != "" {
        cfg.Redactor.RedactionWord = redactionWord
    } 

    var wg sync.WaitGroup
    leaktkScanner := scanner.NewScanner(cfg)
    leaktkRedactor := redactor.NewRedactor(cfg)

    // Note: As more kinds are supported this will be refactored
    // the code below will be moved into the redactor
    if kind != "Stdio" {
        logger.Fatal("unsupported kind: kind=%q", kind)
    }

    go leaktkScanner.Recv(func(response *proto.Response) {
        if len(response.Results) > 0 {
            redactedText, err := leaktkRedactor.RedactText(response.Resource, response)
            if err != nil {
                logger.Error("could not redact text: %v, offset=%s, len=%d", err, response.ID, len(response.Resource))
            } 
            fmt.Print(redactedText)
        }
        wg.Done()
    })

    offset := 0
    err := yieldChunks(cmd.Context(), func(chunk []byte, err error) error {
        if chunkSize := len(chunk); chunkSize > 0 { 
            offset += chunkSize
            wg.Add(1)
            leaktkScanner.Send(&proto.Request{
                ID: strconv.Itoa(offset) ,
                Kind: proto.TextRequestKind,
                Resource: string(chunk),
            })
        }
        time.Sleep(128 * time.Millisecond)
        return err 
    })

    if err != nil && err != io.EOF {
        logger.Fatal("error reading from stdin: %v", err)
    }

    wg.Wait()
}
```

bplaxco (commented):
Added a few comments. What you have looks like it should work, but just wanted to make some structural changes to handle the other redaction stuff that I want to start adding in after this is merged in 👌 (e.g. like being able to pass in results from a previous scan of some files or repo and have it process those instead of doing a new scan, etc)

pkg/redactor/redactor.go
```diff
@@ -0,0 +1,100 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			n, err := r.Read(buf)
+			logger.Info("About to scan")
+			if n > 0 {
+				logger.Info("2")
+				chunk := string(buf[:n])
+				sanitizedChunk, err := s.scanChunk(ctx, detector, chunk, redactionMark, redactionWord)
```
sharman-rh: Important to note that cases where a secret lies on the edge between two separate chunks will not be fully redacted. For example:

A secret is, lets say, 20 bytes long

If 1st 2048-byte chunk end with the first 10 bytes of the secret, and the 2nd 2048-byte chunk starts with the last 10 bytes of the secret, only the first 10 bytes of that secret will be redacted. This applies to secrets larger than 2048 bytes at well, or anything that might overlap with the edge of the chunk.

You can probably fix it by using a sliding redaction window so there's some overlap, or by using context with the previous chunk. Ex: if the 1st chunk ended with a secret on the last character of the chunk, and the 2nd chunk starts with a character like a space (or some other delimiter), continue the redaction until some delimiter is detected.

This also makes a big challenge in trying to use redactionWord. If the secret exists between two chunks, we have to figure out a way to have one word replace two separate redactions.

alayne222: Yeah, that's a good point. Braxton already suggested increasing the chunk size to something like (64 * 1024), so that should hopefully solve the problem of a secret being larger than the chunk size. For the rest of it, we can brainstorm on Monday and hopefully come up with the best way to solve it.

pkg/redactor/redactor.go
```diff
@@ -0,0 +1,54 @@
...
+	mark := r.RedactionMark
+	word := r.RedactionWord
+
+	if len(mark) == 0 && len(word) == 0 {
+		mark = "*"
+	}
+
+	if len(response.Results) == 0 {
+		return resource, nil
+	}
```
bplaxco: nitpick: Just to exit earlier if we don't have any results.

```suggestion
	if len(response.Results) == 0 {
		return resource, nil
	}
	
	mark := r.RedactionMark
	word := r.RedactionWord

	if len(mark) == 0 && len(word) == 0 {
		mark = "*"
	}
```

pkg/redactor/redactor.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		if word != "" {
+			resource = strings.ReplaceAll(resource, result.Secret, word)
+		} else {
+			mask := strings.Repeat(mark, len(result.Secret))
+			resource = strings.ReplaceAll(resource, result.Secret, mask)
+		}
+	}
+	return resource, nil
```
bplaxco: I just remembered why ReplaceAll was bugging me in the back of my mind. If the secret is encoded, I think this would miss it.

What you'll probably want to do is something like [redact-in-place-for](https://github.com/leaktk/hack/blob/main/redact-in-place-for) where it uses start/stop line/columns from `result.Location` and swaps them out.

If you want to minimize allocations during this you could put a [strings.Builder](https://pkg.go.dev/strings#Builder) on the Redactor as a field and run this at the beginning of the function:

```go
r.builder.Reset()
r.builder.Grow(len(resource))
```

and it'll reuse the same underlying memory and grow it to the size of the largest resource we pass in.

Then you can slice up the resource based on the locations and kind of build it back from original parts + the redaction text.

Thoughts?

bplaxco: Or if we want it simpler... we can get rid of `redaction-word` and just byte for byte replace it with redaction mark. Then we could convert the resource to bytes and do all the replacements in place.

If we did that, you might keep a bytes buffer on the redactor to avoid having to allocate byte slices for each string.

bplaxco (changes requested):

cmd/cmd.go
```diff
@@ -341,6 +345,92 @@ func versionCommand() *cobra.Command {
...
+	leaktkScanner := scanner.NewScanner(cfg)
+	leaktkRedactor := redactor.NewRedactor(cfg)
+
+	// Note: As more kinds are supported this will be refactored
+	// the code below will be moved into the redactor
+	if kind != "Stdio" {
+		logger.Fatal("unsupported kind: kind=%q", kind)
+	}
+
+	go leaktkScanner.Recv(func(response *proto.Response) {
```
sharman-rh, line 438: No ordering for concurrent workers. If someone sets their scan_workers in config to be greater than 1, there's a possibility that chunks will be redacted out of order. I tested it with varying numbers of scan_workers in my config.toml and around half of the time there was a chunk out of order.

bplaxco: Good catch. I think a couple of options are:

- Force scan workers to 1 for this (would need to make sure there aren't other gotchas)
- Add a little code to help ensure ordered responses for this case[^1]

If we did the second option, we could just keep track of the request IDs and if it's not the first one that we queued up, we store the response in a `map[string]*Response` until it's ready:

pseudo code example:

```go
// inside the recv loop:
    mutex.Lock()
    backlog[response.RequestID] = response
    
    while response, ok := backlog[requestQueue.Head()]; ok {
        requestQueue.Next() // progress the head to the next one
         // process response here
    }
    mutex.Unlock()
```

A ring buffer or similar could help keep the requestQueue in the example from getting too large over time if we care. I have a ready made ring buffer we could plop into the project if we need it. I was using it for something else.

Also if doing the map + loop every time adds too much delay, you can add a conditional to see if `response.RequestID == requestQueue.Head()` and go ahead and process the response directly instead of dealing with a loop.

(& that's just one thought on how to do it ^_^)

[^1]: The only reason I'm leaning to the second option is there's nothing about the code that makes guarantees for the order, so forcing it to 1 worker may fix it but it's fixed by "accident" if you know what I mean.

cmd/cmd.go
```diff
@@ -390,6 +394,116 @@ func versionCommand() *cobra.Command {
...
+		if chunkSize := len(chunk); chunkSize > 0 {
+			offset += chunkSize
+			wg.Add(1)
+			id := strconv.Itoa(offset)
+			mu.Lock()
+			requestQueue = append(requestQueue, id)
+			mu.Unlock()
+
+			leaktkScanner.Send(&proto.Request{
+				ID:       strconv.Itoa(offset),
```
bplaxco: `id` is already defined above 👍 

```suggestion
				ID:       id,
```

pkg/redactor/redactor.go
```diff
@@ -0,0 +1,137 @@
...
+func computeLineOffsets(s string) []int {
+	offsets := []int{0}
+
+	for i := 0; i < len(s); i++ {
+		if s[i] == '\n' {
+			offsets = append(offsets, i+1)
+		}
+	}
+
+	return offsets
```
bplaxco, line 41: I wanted to see if strings.IndexByte was any faster for stuff like this and it appears so:

```
$ go test -bench=.
goos: linux
goarch: amd64
pkg: offsettest
cpu: Intel(R) Core(TM) Ultra 7 165H
BenchmarkOffsets/Loop-22         	  286335	      3534 ns/op
BenchmarkOffsets/Strings-22      	 1858104	       641.9 ns/o
```
(The Strings one is the suggestion below).

I was curious because of another situation, but figured it was worth the tweak since.

```suggestion
    offsets := []int{0}
    offset := 1
    n := len(s)

    for offset < n {
        i := strings.IndexByte(s[offset:], '\n')
        if i < 0 {
            break
        }
        offset += i + 1
        offsets = append(offsets, offset)
    }

    return offsets
```

pkg/redactor/redactor.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, s := range spans {
+		b.WriteString(resource[cursor:s.Start])
+
+		if r.RedactionWord != "" {
+			b.WriteString(r.RedactionWord)
+		} else {
+			mark := r.RedactionMark
+			if mark == "" {
+				mark = "*"
+			}
```
bplaxco: Lets move the check for setting a redaction mark into config so that we can just always assume the redactor will do what ever the config says. There's a function in config that handles setting missing values 👍

cmd/cmd.go
```diff
@@ -390,6 +394,116 @@ func versionCommand() *cobra.Command {
...
+
+	// Note: As more kinds are supported this will be refactored
+	// the code below will be moved into the redactor
+	if kind != "Stdio" {
+		logger.Fatal("unsupported kind: kind=%q", kind)
+	}
+
+	var mu sync.Mutex
+	backlog := make(map[string]*proto.Response)
+	var requestQueue []string
```
bplaxco, line 435: We _might_ have to keep an eye on how big this gets if someone is using redact on a really large io stream. Or I can plop a little ring buffer that'll wait when it's full into this branch from another thing I've been working on for LeakTK/Betterleaks.

pkg/redactor/redactor.go
```diff
@@ -0,0 +1,137 @@
...
+
+		if s.Start <= last.End {
+			if s.End > last.End {
+				last.End = s.End
+			}
+			continue
+		}
+
+		merged = append(merged, s)
+	}
```
bplaxco, line 75: nitpick that the compiler probably alredy optimizes for you:


```suggestion
    var last Span
    last = merged[0]
    
	for _, s := range spans[1:] {
		if s.Start <= last.End {
			if s.End > last.End {
				last.End = s.End
			}
			continue
		}

		merged = append(merged, s)
		last = s
	}
```

That way you don't have to lookup the last from the list on each loop.

cmd/cmd.go
bplaxco: Moving to another PR because of an issue I ran into with the ring buffer I was going to pull in.

bplaxco (approved):
LGTM!

bplaxco:
Yeah I see what you were talking about with the first character of the match being off from the location. When you spot check it tomorrow as long as it's the scanner results that are off by a little bit and not the redactor picking the wrong region, this looks good to me :+1:

bplaxco:
Yeah from what I can tell, it lines up with what I remember where upstream reports offsets off by one because of the place where the newline is counted (i.e. at the beginning of the next line instead of the end of the previous). 

If your testing confirms the same, I might just work with upstream to see if they'd be willing to tweak it to report column offsets correctly.

bplaxco:
Once you're done with testing, let me know and I'll merge it in and release this bit.

Issue #100, opened 2025-02-17
RFE: `leaktk-scanner grep` or simillar feature


bplaxco:
# Issue:

Some times I need to search a source for an ad-hoc regex. It would be nice if I could use the scanner's integrations with different things to just run a simple regex check against a resource.

# Idea

Either add a `grep` command or a flag to `scan` that takes a regex to use.

Example with stand alone command:

(Benefit: can have custom syntax to make it easier to use)

```sh
leaktk-scanner grep 'secret="[^\"]+"' https://github.com/leaktk/fake-leaks.git
leaktk-scanner grep --kind ConainerImage 'secret="[^\"]+"' quay.io/leaktk/fake-leaks:v1.0.1
leaktk-scanner grep --options '{"since": "2025-01-01"}' 'secret="[^\"]+"' https://github.com/leaktk/fake-leaks.git 

# Usage: leaktk-scanner grep [<optional-scan-flags>] <regex> <resource>
```

Example with flag to scan:

(Benefit: doesn't add an extra command)

```
leaktk-scanner scan --grep 'secret="[^\"]+"' --resource https://github.com/leaktk/fake-leaks.git
leaktk-scanner scan --kind ConainerImage --grep 'secret="[^\"]+"' --resource quay.io/leaktk/fake-leaks:v1.0.1
leaktk-scanner scan --options '{"since": "2025-01-01"}' --grep 'secret="[^\"]+"' --resource https://github.com/leaktk/fake-leaks.git 

# Usage: leaktk-scanner scan [<optional-scan-flags>] [--grep <regex>] --resource <resource>
```

Thoughts?

Aside: We could also take scan and make the `--resource` flag a positional argument since it's always required. It'd make scanning remote git repos as easy as: `leaktk-scanner scan <repo-url>`.

rgmz:
What's the difference between this and a typical scan command? Being able to pass-in an arbitrary pattern?

bplaxco:
The main difference would be only invoking gitleaks as a backend (or a custom grep backend).

Right now we have one backend, but I have yara support in my backlog.

For grep, I think I'd want the same result format that everything returns now, but just using that one regex.

For the rule field I could leave that empty or do the ID as a hash of the regex and the description be "Custom Regex"

bplaxco:
I'm often hunting for random leak related things and testing patterns and this would be very handy for that.

bplaxco:
Done! Scan has a `--grep` flag now.

PR #278, opened 2026-07-07
Fixing issue #100 - adding grep scan functionality

sharman-rh:
Added logic to `scan` to allow for ad-hoc regex to be used during scans by adding a `--grep` flag.

Usage example: `leaktk scan --grep 'secret="[^\"]+"' --resource https://github.com/leaktk/fake-leaks.git`

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -8,10 +8,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"slices"
 	"strings"
 	"sync"
-	"regexp"
 
 	"github.com/spf13/cobra"
 
@@ -107,13 +107,13 @@ func runScan(cmd *cobra.Command, args []string) {
 		if err != nil {
 			logger.Fatal("could not create a temp config: %v", err)
 		}
-		defer os.Remove(tmpFile.Name())
+		defer func() { _ = os.Remove(tmpFile.Name()) }()
 
 		if _, err := tmpFile.WriteString(buildGitleaksConfig(grepPattern)); err != nil {
-			tmpFile.Close()
+			_ = tmpFile.Close()
 			logger.Fatal("could not write temp config: %v", err)
 		}
-		tmpFile.Close()
+		_ = tmpFile.Close()
 
 		gitleaksConfig = tmpFile.Name()
 	}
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -255,10 +281,17 @@ func scanCommand() *cobra.Command {
 	flags.StringP("options", "o", "{}", "Provide scan specific options formatted as JSON")
 	flags.Int("leak-exit-code", 0, "Exit with this code when leaks are detected (default 0)")
 	flags.String("gitleaks-config", "", "Load a custom gitleaks config")
+	flags.StringP("grep", "g", "", "Scan using ad-hoc regex instead of the configured patterns")
 
 	return scanCommand
```
bplaxco, line 290: [This](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) will help keep people from combining `--grep` and `--gitleaks-config`.

```suggestion
	// Ensure incompatible flags can't be combined
	scanCommand.MarkFlagsMutuallyExclusive("grep", "gitleaks-config")

	return scanCommand
```

cmd/cmd.go
```diff
@@ -255,10 +281,17 @@ func scanCommand() *cobra.Command {
...
 }
 
+func buildGitleaksConfig(pattern string) string {
+	ruleID := id.ID(pattern)
+
+	return fmt.Sprintf("[[rules]]\nid = %q\ndescription = \"Custom Regex\"\nregex = '''%s'''\n", ruleID, pattern)
```
bplaxco: To avoid potential quoting issues and such, might do something like this below. It's a bit more to do but should handle most of the annoying edge cases and such.


```suggestion
func buildGitleaksConfig(pattern string) string {
        var rule struct {
                ID          string `toml:"id"`
                Description string `toml:"description"`
                Regex       string `toml:"regex"`
        }
        rule.ID = id.ID("grep", pattern)
        rule.Description = "Grep Pattern Match"
        rule.Regex = pattern

        data, err := toml.Marshal(&rule)
        if err != nil {
                logger.Fatal("could not marshal grep rule: %v", err)
        }

        return fmt.Sprintf("[[rules]]\n%s\n", string(data))
```

sharman-rh: That does seem like a much better idea so stuff doesn't break because of certain characters.

bplaxco:
Noice, tested it and it works:


```
./leaktk scan --grep '-----BEGIN [^-]+-----[\s\S]+-----END [^-]+-----' https://github.com/leaktk/fake-leaks.git | jq
```

Issue #272, opened 2026-06-30
[RFE] Hook `posix.stdio`


bplaxco:
# Overview

This is part of the larger [hook's
feature](https://github.com/leaktk/leaktk/issues/238) and focuses on shell
stdio. It replaces stdout and stderr file descriptors and filters the output
redacting found secrets

## Use Case

Mainly CI/CD systems where stdout and stderr (especially because of `set -x`)
can leak secrets to the logs. But could also be useful for doing recordings and
other things where having your stdout redacted may be useful.

## Requirements

Each component is broken out to its own sections and steps out the
implementation:

### Hook Installer

```
USAGE
   leaktk hook install posix.stdio [install-flags]

DESCRIPTION
    Adds this line to the specified target if it doesn't exist:

        eval "$(leaktk hook posix.stdio)"

    It checks if it exists by searching for the regex:

        \bleaktk\s+hook\s+posix.stdio\b

    The following install-flags set the target:

        --bashrc      Targets ~/.bashrc

        --zshrc       Targets ~/.zshrc

        --stdout      Prints the command to stdout for easier ad-hoc custom
                      installs
```

### Hook

```
USAGE

    leaktk hook posix.stdio

DESCRIPTION
    Prints out commands to hook into the current shell session. It is meant
    to be eval'd.

    It should output the POSIX equivalent to the following bash:

      (See: "Potential `leaktk hook posix.stdio` output script:" below)

NOTES
    We may need to play around a bit with how the redirection works and 
    handle edge cases and we may want to look at how to handle interactive 
    terminals at some point but lets get something like this going to start.
```

### Redact 

```
USAGE
    leaktk redact --kind Stdio

DESCRIPTION
    Reads data in chunks on stdin, redacts found secrets, and prints it to 
    stdout.

    MUST
        Replace the secret with substitute characters (i.e. '*')

    SHOULD
        Keep data flowing and not wait for a huge buffer to fill up. See notes
        on buffering below.

    MAY
        Have these optional, mutually exclusive flags: 
            --redaction-mark    replace each rune of the secret with this rune (default: *)
            --redaction-word    repalce each whole secret with this word (no default)

BUFFERING
    We don't want to wait until large chunks of text fill an entire buffer and
    we also don't want to only read it line by line because of multiline
    secrets. It would be nice to read UP TO a certain number of bytes that are
    available (where it's okay to read less if it's not available). Probably
    with a delay between reads of around 128ms to give things keys being
    generated time to print out.
```

[...]

PR #282, opened 2026-07-23
Hook posix stdio

alayne222:
Creates hook explained in #272 using the new redact command.


cmd/install.go
```diff
@@ -54,6 +56,22 @@ func gitHookInstallCommand(hook hooks.Hook) *cobra.Command {
...
+		Use:   hookname,
+		Short: "Install and configure " + hookname,
+		Run:   runPosixHookInstall,
+	}
+	flags := cmd.Flags()
+	var bashrc, zshrc, stdout bool
+	flags.BoolVar(&bashrc, "bashrc", false, "Target ~/.bashrc for installation")
+	flags.BoolVar(&zshrc, "zshrc", false, "Target ~/.zshrc for installation")
+	flags.BoolVar(&stdout, "stdout", false, "Print the command to stdout for ad-hoc custom installs")
+	cmd.MarkFlagsMutuallyExclusive("bashrc", "zshrc", "stdout")
```
bplaxco: @alayne222 Would there be any issues with making it not mutually exclusive? I'm thinking it might be handy to support multiple targets for the install command like the `git.*` hooks do.

pkg/hooks/posix_stdio.go
```diff
@@ -0,0 +1,52 @@
...
+	# keep the pipes active until the background processes terminate.
+	rm    "${_leaktk_hook_posix_stdio_stdout_fifo}"
+	rm    "${_leaktk_hook_posix_stdio_stderr_fifo}"
+	rmdir "${_leaktk_hook_posix_stdio_tmpdir}"
+
+	# Unset environment cleanup variables
+	unset _leaktk_hook_posix_stdio_tmpdir
+	unset _leaktk_hook_posix_stdio_stdout_fifo
+	unset _leaktk_hook_posix_stdio_stderr_fifo
+	`
```
bplaxco: Thoughts on moving this up as `const posixStdioHookRunScript`?

Rational:

- No need for indentation on the left (and the trim space can be removed)
- Letting the compiler know it's a const
- Might be handy for some test cases in the future

pkg/hooks/posix_stdio.go
```diff
@@ -0,0 +1,52 @@
...
+	_, err := fmt.Fprint(os.Stdout, strings.TrimSpace(hookScript))
```
bplaxco: Since this is to stdout, would `fmt.Print` work?

pkg/installer/posix_stdio.go
```diff
@@ -0,0 +1,92 @@
...
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("could not detect user home directory: %w", err)
+	}
+	targetPath := filepath.Join(homeDir, targetFilename)
+
+	file, err := os.OpenFile(targetPath, os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return fmt.Errorf("faile to open or create target file %s: %w", targetPath, err)
```
bplaxco:
```suggestion
		return fmt.Errorf("failed to open or create target file: %w path=%q", err, targetPath)
```

- typo
- providing variable data as a key=value pair like some other log places.

pkg/installer/posix_stdio.go
```diff
@@ -0,0 +1,92 @@
...
+	defer func() {
+		if closeErr := file.Close(); closeErr != nil && err == nil {
+			err = closeErr
+		}
+	}()
+
+	hookRegex, err := regexp.Compile(`\bleaktk\s+hook\s+posix\.stdio\b`)
+	if err != nil {
+		return fmt.Errorf("failed to compile idempotency regex: %w", err)
+	}
```
bplaxco: Since the regex should always compile, you could do a MustCompile and save it as var at the top of the file. (Then if we remove the mutually exclusive check on the install targets, the function won't need to recompile it for each pass if you implement the common install steps in their own func).

pkg/installer/posix_stdio.go
```diff
@@ -0,0 +1,94 @@
...
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+const hookEvalLine = `eval "$(leaktk hook posix.stdio)"`
```
bplaxco:
```suggestion
const posixStdioHookEvalLine = `eval "$(leaktk hook posix.stdio)"`
```

Since this is in the installer package with other hooks, we'll want to namespace the variable.

pkg/installer/posix_stdio.go
bplaxco, line 1: nitpick:

Filename `posix_stdio.go` -> `posix_stdio_hook.go`

Just since the prefix for all the things here is `PosixStdioHook`.

pkg/installer/posix_stdio.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to compile idempotency regex: %w", err)
+	}
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if hookRegex.MatchString(scanner.Text()) {
+			return nil
+		}
+	}
```
bplaxco: Would it make sense to do something like:


```suggestion
	if len(hookRegex.FindReaderIndex(bufio.NewReader(file))) > 0 {
	    return nil
	}
```

pkg/installer/posix_stdio.go
```diff
@@ -0,0 +1,94 @@
...
+			appendPrefix = "\n"
+		}
+	}
+	if _, err = file.Seek(0, io.SeekEnd); err != nil {
+		return fmt.Errorf("failed to seek to end of file: %w", err)
+	}
+
+	if _, err = fmt.Fprintf(file, "%s%s\n", appendPrefix, hookEvalLine); err != nil {
+		return fmt.Errorf("failed to write hook to configuration file: %w", err)
+	}
```
bplaxco: Would this cause any issues?


```suggestion
	if _, err = file.Seek(0, io.SeekEnd); err != nil {
		return fmt.Errorf("failed to seek to end of file: %w", err)
	}
	if _, err = fmt.Fprintf(file, "\n%s\n", hookEvalLine); err != nil {
		return fmt.Errorf("failed to write hook to configuration file: %w", err)
	}
```

I figure an extra newline before would look good stylistically in most cases and functionally in cases where the file didn't end in a newline (which it should most of the time 👌 )

pkg/installer/posix_stdio.go
bplaxco: Also we should probably add a logger statment like:

```go
logger.Info("skipping file: posix.stdio hook reference detected path=%q", targetPath)
```

pkg/hooks/posix_stdio.go
alayne222: I also think we can simplify the redact command lines to :
`
leaktk redact --kind Stdio < "${_leaktk_hook_posix_stdio_stdout_fifo}" 2>"${LEAKTK_LOGGER_FILE}" &
leaktk redact --kind Stdio < "${_leaktk_hook_posix_stdio_stderr_fifo}" 2>"${LEAKTK_LOGGER_FILE}" &
`

Although I'm not sure how beneificial that is other than simplification

cmd/install.go
```diff
@@ -54,6 +56,21 @@ func gitHookInstallCommand(hook hooks.Hook) *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:   hookname,
+		Short: "Install and configure " + hookname,
+		Run:   runPosixHookInstall,
+	}
+	flags := cmd.Flags()
+	var bashrc, zshrc, stdout bool
+	flags.BoolVar(&bashrc, "bashrc", false, "Target ~/.bashrc for installation")
+	flags.BoolVar(&zshrc, "zshrc", false, "Target ~/.zshrc for installation")
+	flags.BoolVar(&stdout, "stdout", false, "Print the command to stdout for ad-hoc custom installs")
```
bplaxco: I just noticed this is the `*Var` variant. Does [flags.Bool](https://pkg.go.dev/github.com/spf13/pflag#Bool) work?


```suggestion

	flags.Bool("bashrc", false, "Target ~/.bashrc for installation")
	flags.Bool("zshrc", false, "Target ~/.zshrc for installation")
	flags.Bool("stdout", false, "Print the command to stdout for ad-hoc custom installs")
```

pkg/hooks/posix_stdio.go
```diff
@@ -0,0 +1,48 @@
...
+# Cleanup files from disk immediately. The open file descriptors in memory 
+# keep the pipes active until the background processes terminate.
+rm    "${_leaktk_hook_posix_stdio_stdout_fifo}"
+rm    "${_leaktk_hook_posix_stdio_stderr_fifo}"
+rmdir "${_leaktk_hook_posix_stdio_tmpdir}"
+
+# Unset environment cleanup variables
+unset _leaktk_hook_posix_stdio_tmpdir
+unset _leaktk_hook_posix_stdio_stdout_fifo
+unset _leaktk_hook_posix_stdio_stderr_fifo`
```
bplaxco: You might want that newline on the end since it's piped out to a shell. It wouldn't fail but I'm thinking it might not run that last unset in certain cases.

And I only moved the comment on the first line down to make it a little nicer to read there but also it's handy if for whatever reason the thing it's being tacked onto (for the stdout cases) doesn't already end in a newline.

```suggestion
const posixStdioHookRunScript = `
# Set user configurable var defaults
: "${LEAKTK_LOGGER_FILE:="/dev/null"}"

# Define tmpdir and fifos using the current shell PID ($$)
_leaktk_hook_posix_stdio_tmpdir="${TMPDIR:-"/tmp"}/leaktk.$$"
_leaktk_hook_posix_stdio_stdout_fifo="${_leaktk_hook_posix_stdio_tmpdir}/stdout"
_leaktk_hook_posix_stdio_stderr_fifo="${_leaktk_hook_posix_stdio_tmpdir}/stderr"

# Create secure directory and named pipes
mkdir -p -m 0700 "${_leaktk_hook_posix_stdio_tmpdir}"
mkfifo           "${_leaktk_hook_posix_stdio_stdout_fifo}"
mkfifo           "${_leaktk_hook_posix_stdio_stderr_fifo}"

# Start background redaction daemons to listen to the pipes
leaktk redact --kind Stdio < "${_leaktk_hook_posix_stdio_stdout_fifo}" 1>&1 2>"${LEAKTK_LOGGER_FILE}" &
leaktk redact --kind Stdio < "${_leaktk_hook_posix_stdio_stderr_fifo}" 1>&2 2>"${LEAKTK_LOGGER_FILE}" &

# Redirect stdout and stderr
exec 1> "${_leaktk_hook_posix_stdio_stdout_fifo}"
exec 2> "${_leaktk_hook_posix_stdio_stderr_fifo}"

# Cleanup files from disk immediately. The open file descriptors in memory 
# keep the pipes active until the background processes terminate.
rm    "${_leaktk_hook_posix_stdio_stdout_fifo}"
rm    "${_leaktk_hook_posix_stdio_stderr_fifo}"
rmdir "${_leaktk_hook_posix_stdio_tmpdir}"

# Unset environment cleanup variables
unset _leaktk_hook_posix_stdio_tmpdir
unset _leaktk_hook_posix_stdio_stdout_fifo
unset _leaktk_hook_posix_stdio_stderr_fifo
`
```

pkg/installer/posix_stdio_hook.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+const posixStdioHookEvalLine = `eval "$(leaktk hook posix.stdio)"`
+
+var hookRegex = regexp.MustCompile(`\bleaktk\s+hook\s+posix\.stdio\b`)
```
bplaxco:
```suggestion
var posixStdioHookRegex = regexp.MustCompile(`\bleaktk\s+hook\s+posix\.stdio\b`)
```

needs to be namespaced

pkg/installer/posix_stdio_hook.go
```diff
@@ -0,0 +1,78 @@
...
+	file, err := os.OpenFile(cleanPath, os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return fmt.Errorf("failed to open or create target file: %w path=%q", err, cleanPath)
+	}
+
+	defer func() {
+		if closeErr := file.Close(); closeErr != nil && err == nil {
+			err = closeErr
+		}
+	}()
```
bplaxco: Assigning to err there won't pass it on to the caller since each return statement below wraps error and the final one returns nil.

If we just want to throw away the error we could do this:


```suggestion
	defer func() { _ = file.Close() }()
```

If we want to return the error, you'd probably want to change the func signature to:

```go
appendToTargetFile(targetFile string, homeDir string) (err error)
```

And that last return either to just `return` or `return err`. And I think that should handle getting the error back to the caller.

pkg/installer/posix_stdio_hook.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if len(hookRegex.FindReaderIndex(bufio.NewReader(file))) > 0 {
+		return nil
```
bplaxco, line 64: See logger comment: https://github.com/leaktk/leaktk/pull/282#discussion_r3639853838

pkg/installer/posix_stdio_hook.go
```diff
@@ -0,0 +1,78 @@
...
+	if opts.Zshrc {
+		if err = appendToTargetFile(".zshrc", homeDir); err != nil {
+			return fmt.Errorf("failed to write hook to .zshrc: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func appendToTargetFile(targetFile string, homeDir string) error {
```
bplaxco: The nitiest of nit picks:


```suggestion
func appendToTargetFile(homeDir, targetFile string) error {
```

Rationale: it just feels backwards in a way that it's hard to explain (note: the calls would need to be updated). If you disagree, feel free to also just resolve this without a change 👌

bplaxco: Also if you want, the first `string` can be removed since both args are strings 👌

bplaxco (commented):
Just a couple of small things. But lookin' good!

pkg/hooks/posix_stdio.go
bplaxco: Yeah, `1>&1` is superfluous but is mainly there to keep it consistent with `1>&2`, but `1>&2` is there so that the stderr from the shell and commands continue to flow to stderr instead of combining the two to stdout.

bplaxco:
LGTM! Probably want to write up some docs and mention in the docs that this is best used in scripts to avoid secrets making it into the logs (with an example of putting `eval "$(leaktk hook posix.stdio)"` right after the shebang). And the main use case for `leaktk install hook posix.stdio` is for cases where it's hard to control every script and it can be ran as a setup step a CI/CD runner for the job.

PR #284, opened 2026-07-24
LeakTK Container: Set git safe.directory = *

bplaxco:
Similar to this but sets it for any user in the container:

https://github.com/betterleaks/betterleaks/blob/main/Dockerfile#L10C1-L11C49


abutcher (approved):
LGTM

Issue #279, opened 2026-07-10
[RFE] Improve login experience


bplaxco:
# Overview

Today the login flow for custom pattern servers is pretty manual and forces users to copy/paste values. Here's an idea to potentially improve that flow. This is modeled after the `oc login` command's behavior.

# Prerequisites

Update existing pattern server to return the proper headers and handle the callback.

# Idea

### Args

Support the following:

```
leaktk login [URL] [flags]
```

where the URL is the server url and the extra flags are:

```
--token=''             Bearer token for authentication to the server.

-w, --web=false        Login with web browser. Starts a local HTTP callback server to perform the OAuth2 Authorization Code Grant
                       flow. Use with caution on multi-user systems, as the server's port will be open to all users.
```

### Web login flow

1. Make a normal request to the pattern server
2. Check the response code
3. If the response code is a 401:
    1. Check for a `WWW-Authenticate` header (in the same way that OCI clients do)
    2. If a WWW-Authenticate header doesn't exist, then raise an authentication error and stop
    3. If it's an OAuth flow, do a browser based flow similar to how the [`oc` command does it](https://github.com/openshift/oc/blob/22c69c0ff5da7e390b42880141e43d555e38ef41/pkg/cli/login/loginoptions.go#L324-L342) with a local callback to handle the token:
        1. If a browser can be opened, do a browser based flow else provide instructions in the terminal for opening a browser
    4. Make another request to the pattern server and confirm the token is good; else error out
6. Store the token in the same place that it's normally stored in (or even the keyring if that's not hard to support).

**To consider/maybe new issue**:

In cases where a browser isn't available and a local server doesn't work (e.g. on a server somewhere), it would be nice to support [Device Authorization Flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/device-authorization-flow) where the user can enter a pin into the browser and get a login token that way. If we do that, we  might now want to implement this at the pattern server level (or maybe we do depending on how complex it is).

## Normal usage changes

Provide a setting for `autologin` at the same place in the config as `autofetch` and also respect a `LEAKTK_AUTOLOGIN` env var. It should default to false so we don't impact hooks and other use cases.

If it's set to true, then given custom pattern server URL + a 401 response, kick off the login flow in a way that won't infinite loop. 

If it's set to false, then display the command needed to log in again and exit with an error code.

(We can set autologin to true as a default for endpoints so they can use the tool without needing to manage updating their tokens from time to time).

---

[...]

PR #287, opened 2026-07-28
Improved login experience

sharman-rh:
Fixing issue #279 

**Web Login Flow**

1. Makes a request to the pattern server. Challenge() sends a GET to the server URL
2. Checks the response code:
- 200: server doesn't require auth, returns an error saying so
- 401: checks for WWW-Authenticate: Bearer header, parses realm/service/scope
- 401 without a usable header: returns "use --token instead" error
- 3xx: follows redirects looking for an OAuth authorization URL (response_type=code)
3. Does OIDC discovery on the realm. Gets authorization and token endpoints
4. Starts a local callback server on 127.0.0.1 with a random port
5. Opens a browser. If it fails, prints the URL to the terminal for manual opening
6. Waits for the callback with the authorization code
7. Exchanges the code for a token via the token endpoint (with PKCE)
8. Validates the token and makes another request to the pattern server with the Bearer token, confirms 200
9. Stores the token to ~/.config/leaktk/pattern-server-auth-token

**What the Pattern Server Needs**

1. Return 401 with a `WWW-Authenticate: Bearer realm="<issuer_url>"` header whenever there isn't a present valid token. The realm should point to OICD provider.
2. Validate `Authorization: Bearer <token>` headers and return 200 when token is good

**What the OIDC Provider needs**

Register a client:
- Client ID: leaktk-cli
- Client type: public
- Grant type: authorization code
- PKCE: S256 required
- Redirect URIs: http://127.0.0.1:*
- Scope: openid


sharman-rh:
Not entirely sure if security errors are FPs or not. The "user-provided value" would be the response from the server, which the client is trusting to be legitimate.

cmd/cmd.go
```diff
@@ -45,20 +45,57 @@ func runHelp(cmd *cobra.Command, args []string) {
...
 
-	fmt.Printf("Enter %s auth token: ", cfg.Scanner.Patterns.Server.URL)
+	logger.Info("logging in: pattern_server=%q", serverURL)
 
-	var authToken string
-	if _, err := fmt.Scanln(&authToken); err != nil {
-		logger.Fatal("could not login: %v", err)
-	}
+	token, _ := cmd.Flags().GetString("token")
+	web, _ := cmd.Flags().GetBool("web")
```
bplaxco:
```suggestion
	flags := cmd.Flags()
	token := mustGetString(flags, "token")
	web := mustGetBool(flags, "bool")
```

Just swapping it out with a couple of helper funcs that we have for these that require the get to work.

cmd/cmd.go
```diff
@@ -45,20 +45,57 @@ func runHelp(cmd *cobra.Command, args []string) {
 }
 
 func runLogin(cmd *cobra.Command, args []string) {
-	logger.Info("logging in: pattern_server=%q", cfg.Scanner.Patterns.Server.URL)
+	serverURL := cfg.Scanner.Patterns.Server.URL
+	if len(args) > 0 {
+		serverURL = args[0]
```
bplaxco, line 60: If they provide a custom serverURL, we'll want to persist that if there's a valid login.

What are your thoughts on:

1. Change `SavePatternServerAuthToken` to `SavePatternServerAuth`
2. Update the signature to be `SavePatternServerAuth(serverURL, authToken string) error`
3. Write the URL to `pattern-server-url` next to `pattern-server-auth-token`
4. If they exist, `pattern-server-url` and `pattern-server-auth-token` take precedence over the values in the config.toml

pkg/auth/auth.go
```diff
@@ -0,0 +1,467 @@
...
+	Scope    string
+	ClientID string
+}
+
+type OIDCEndpoints struct {
+	AuthorizationEndpoint string `json:"authorization_endpoint"`
+	TokenEndpoint         string `json:"token_endpoint"`
+}
+
+func ParseWWWAuthenticate(header string) (WWWAuthenticate, error) {
```
bplaxco: No need to change, just sharing a neat thing.

If this was a method on `WWWAuthenticate` called [UnmarshalText](https://pkg.go.dev/encoding#TextUnmarshaler), it'll work the same here but also provide some nifty features with other encoders/decoders if ever needed.


To use it, you'd end up basically:

```
var auth WWWAuthenticate

if err := auth.UnmarshalText(header); err != nil {
   ...
}
```

But then if it's ever a field on something else and we unmarshal it, this auto detects that the function is present and magically maps things 👌

pkg/auth/auth.go
```diff
@@ -0,0 +1,467 @@
...
+			current.WriteRune(r)
+		}
+	}
+
+	if current.Len() > 0 {
+		parts = append(parts, current.String())
+	}
+
+	return parts
+}
```
bplaxco: Would something like this work?

Defined at the top of the file

```
var wwwAuthParamRe = regexp.MustCompile(`([^\s"=,]+)\s*=\s*(?:"((?:[^"\\]|\\.)*)"|([^\s",]+))`)
```

```suggestion

func parsePrams(s string) map[string]string {
	params := make(map[string]string)
	for _, match := range wwwAuthParamRe.FindAllStringSubmatch(s, -1) {
		key := match[1]
		if match[2] != "" {
			// It matched the quoted group (Group 2)
			// Clean up escaped quotes (convert \" back to ")
			params[key] = strings.ReplaceAll(match[2], `\"`, `"`)
		} else {
			// It matched the unquoted group (Group 3)
			params[key] = match[3]
		}
	}
	return params
}
```

Then you should have a nice dict of params to pull from 👌

pkg/auth/auth.go
```diff
@@ -0,0 +1,467 @@
...
+
+func DiscoverEndpoints(ctx context.Context, client *http.Client, realm string) (*OIDCEndpoints, error) {
+	discoveryURL := strings.TrimRight(realm, "/") + "/.well-known/openid-configuration"
+
+	request, err := http.NewRequestWithContext(ctx, "GET", discoveryURL, nil)
+	if err != nil {
+		return fallbackEndpoints(realm), nil
+	}
+
+	response, err := client.Do(request)
```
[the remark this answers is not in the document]
bplaxco: @puiterwijk do you have any thoughts on proper handling here, libs to handle it properly, etc? I figured you'd be a great person to ping on this one.

pkg/auth/auth.go
```diff
@@ -0,0 +1,467 @@
...
+	if err != nil {
+		return fallbackEndpoints(realm), nil
```
bplaxco: I think, I'd just not have it fall back on anything unless there's some part of the OIDC spec that covers what fallbacks should be if one isn't discovered.

pkg/auth/browser_windows.go
```diff
@@ -0,0 +1,9 @@
+//go:build windows
+
+package auth
+
+import "os/exec"
+
+func OpenBrowser(url string) error {
```
bplaxco: Thoughts on using github.com/pkg/browser (just worried about dragons with exec commands from web input)

pkg/auth/auth.go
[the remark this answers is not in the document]
bplaxco: @sharman-rh or actually it might just make more sense to pull in github.com/coreos/go-oidc/v3 ([similar to how oc does](https://github.com/openshift/oc/blob/a4c90c71ea5562102352df426c71e03381b7a02f/pkg/cli/gettoken/gettoken.go)).

That should handle a lot of the sharp edges here.

bplaxco (changes requested):

bplaxco:
> Not entirely sure if security errors are FPs or not. The "user-provided value" would be the response from the server, which the client is trusting to be legitimate.

I think the switch to an oidc client should help (https://github.com/leaktk/leaktk/pull/287/changes#r3683688135)

cmd/cmd.go
sharman-rh: I like that! I just implemented it and it works great. I'll push in a few

pkg/auth/browser_windows.go
sharman-rh: Not a problem, that seems much safer

pkg/scanner/patterns.go
```diff
@@ -199,3 +213,40 @@ func (p *Patterns) Gitleaks(ctx context.Context) (*betterleaksconfig.Config, err
...
+	if !errors.Is(fetchErr, ErrUnauthorized) {
+		return "", fetchErr
+	}
+
+	if !p.config.Autologin || !p.isCustomServer() {
+		return "", fmt.Errorf(
+			"authentication required: run \"leaktk login %s\" to authenticate: %w",
+			p.config.Server.URL, fetchErr,
+		)
+	}
```
bplaxco: I don't think the patterns should "know" or "care" if they're using a custom server. Instead, `AutoLogin` should be set in the config package based on that logic.

For example if in the future we change the default server to one that required login (not saying we would, but just as an example), then this would need to be changed too. If we keep the logic in Config, we only have to update it in one place and the rest of the code would just read the flags it was provided and wouldn't have to "know" about that change.

sensitive-file.txt
bplaxco, line 1: I know it's a fake JWT, but it shouldn't be checked in here. Can you delete it?

pkg/auth/callback.go
```diff
@@ -0,0 +1,85 @@
...
+		_ = server.Serve(listener)
+	}()
+
+	shutdown := func() {
+		_ = server.Shutdown(context.Background())
+	}
+
+	go func() {
+		<-ctx.Done()
+		shutdown()
```
bplaxco: `StartCallbackServer` spawns a goroutine that calls `shutdown()` on context cancellation, and `WebLogin` also defers `shutdown()`. The `http.Server.Shutdown` docs say calling it multiple times is safe, so this isn't a bug, but the goroutine is unnecessary if the caller always defers `shutdown()`. Removing the goroutine simplifies the lifecycle and avoids any ambiguity about who owns shutdown.

pkg/scanner/patterns.go
```diff
@@ -199,3 +213,40 @@ func (p *Patterns) Gitleaks(ctx context.Context) (*betterleaksconfig.Config, err
...
+
+	logger.Info("authentication required, attempting autologin: server=%q", p.config.Server.URL)
+
+	token, err := auth.WebLogin(ctx, p.client, p.config.Server.URL)
```
bplaxco, line 230: If `autologin=true` is set but the process is running in a CI pipeline or cron job (no display), this will fail or hang. The `isCustomServer()` gate is good but a TTY check (e.g. `term.IsTerminal(int(os.Stdin.Fd()))`) before attempting browser login would be a safer guard.

sensitive-file.txt
sharman-rh: My bad, I created a fake one for a demo in my expo presentation (oops). I can remove it

pkg/auth/auth.go
```diff
@@ -0,0 +1,437 @@
...
+	endpoint := provider.Endpoint()
+
+	if len(endpoint.AuthURL) == 0 || len(endpoint.TokenURL) == 0 {
+		return nil, errors.New("OIDC discovery response missing required endpoints")
+	}
+
+	return &OIDCEndpoints{
+		AuthorizationEndpoint: endpoint.AuthURL,
+		TokenEndpoint:         endpoint.TokenURL,
+	}, nil
```
bplaxco: nitpik:

This `OIDCEndpoints` struct is only used to return two strings from `DiscoverEndpoints`. Since `go-oidc` gives you `provider.Endpoint()`, you could `return endpoint, nil` instead and not need an extra struct.

pkg/auth/auth.go
```diff
@@ -0,0 +1,437 @@
...
+func noRedirectClient(client *http.Client) *http.Client {
+	return &http.Client{
+		Transport: client.Transport,
+		Timeout:   client.Timeout,
+		Jar:       client.Jar,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+	}
+}
```
bplaxco, line 142:
```suggestion
func noRedirectClient(client *http.Client) *http.Client {
	c := *client
	c.CheckRedirect = func(req *http.Request, via []*http.Request) error {
			return http.ErrUseLastResponse
		},
	}
	return &c
}
```

This ensures all of the fields from the client are copied over (e.g. user agent, etc)

cmd/cmd.go
```diff
@@ -45,38 +45,82 @@ func runHelp(cmd *cobra.Command, args []string) {
...
+		}
+
+		if err := config.SavePatternServerAuth(serverURL, authToken); err != nil {
+			logger.Fatal("could not save token: %v", err)
+		}
 	}
 
-	logger.Info("token saved")
+	logger.Info("login successful")
 }
```
bplaxco, line 92: This duplicates validate-then-save across all three branches

The `--token`, `--web`, and interactive cases in `cmd.go` each independently call `ValidateToken` + `SavePatternServerAuth`. Extract a small helper to reduce the repetition:

```go
func saveLogin(ctx context.Context, client *http.Client, serverURL, token string) {
    if err := auth.ValidateToken(ctx, client, serverURL, token); err != nil {
        logger.Fatal("token validation failed: %v", err)
    }
    if err := config.SavePatternServerAuth(serverURL, token); err != nil {
        logger.Fatal("could not save token: %v", err)
    }
}
```

pkg/auth/auth.go
bplaxco, line 1: ## Refactor: use `golang.org/x/oauth2` for the OAuth flow

The PR already pulls in `go-oidc`, which is designed to pair with `golang.org/x/oauth2`. Right now the code hand-rolls PKCE generation, auth URL construction, and token exchange. All of that can be replaced with library calls, which removes ~100 lines and their tests.

Here's the shape of the change:

**Delete:** `GeneratePKCE`, `PKCEChallenge`, `BuildAuthURL`, `ExchangeCode`, `OIDCEndpoints` struct, and their tests.

**Replace `DiscoverEndpoints`:** Return an `oauth2.Config` directly.

```go
func OAuthConfig(ctx context.Context, client *http.Client, realm, clientID, redirectURI string) (*oauth2.Config, error) {
    if len(clientID) == 0 {
        clientID = defaultClientID
    }
    ctx = oidc.ClientContext(ctx, client)
    provider, err := oidc.NewProvider(ctx, realm)
    if err != nil {
        return nil, fmt.Errorf("OIDC discovery failed: %w", err)
    }
    return &oauth2.Config{
        ClientID:    clientID,
        Endpoint:    provider.Endpoint(),
        RedirectURL: redirectURI,
        Scopes:      []string{oidc.ScopeOpenID},
    }, nil
}
```

**Simplify `WebLogin`:** Use `oauth2.GenerateVerifier()` for PKCE and `oauth2.Config` methods for the URL and exchange.

```go
verifier := oauth2.GenerateVerifier()
authURL := cfg.AuthCodeURL(state, oauth2.S256ChallengeOption(verifier))

// ... open browser, wait for callback ...

token, err := cfg.Exchange(ctx, result.Code, oauth2.VerifierOption(verifier))
if err != nil {
    return "", err
}
return token.AccessToken, nil
```

**What stays as-is:** `ParseWWWAuthenticate` / `parseParams` (no library covers this), `Challenge` + `discoverFromRedirects` (application-specific), `ValidateToken` (application-specific), `StartCallbackServer` (application-specific).

pkg/auth/callback.go
sharman-rh: Gotcha, so I can just remove that?

pkg/scanner/patterns.go
sharman-rh: If it were to run on a CI pipeline, the browser login wouldn't even be attempted though, correct?

sharman-rh: That's a simple enough check though, better safe than sorry

cmd/testserver/main.go
bplaxco, line 1: I haven't tested it, but here's a little script we could yeet in `hack/mock-pattern-server` until the pattern server code is moved over here. (Disclaimer: AI generated) but figured it'd help keep the testing infra a little more separate from the project.

Should be able to test it via:

```
leaktk login http://127.0.0.1:8580 --web
```

Thoughts?

```python
#!/usr/bin/env python3

import shutil
import signal
import subprocess
import sys
import time
from http.server import HTTPServer, BaseHTTPRequestHandler
from urllib.error import HTTPError, URLError
from urllib.request import Request, urlopen

DEX_HOST = "127.0.0.1"
DEX_PORT = 5556
DEX_ISSUER = f"http://{DEX_HOST}:{DEX_PORT}/dex"

SERVER_HOST = "127.0.0.1"
SERVER_PORT = 8580

DEX_CONFIG = f"""\
issuer: {DEX_ISSUER}
storage:
  type: memory
web:
  http: {DEX_HOST}:{DEX_PORT}
staticClients:
- id: leaktk-cli
  name: LeakTK CLI
  public: true
  redirectURIs:
  - http://127.0.0.1/callback
enablePasswordDB: true
staticPasswords:
- email: admin@example.com
  hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
  username: admin
"""

GITLEAKS_CONFIG = """
[[rules]]
id = "mock-secret"
description = "Mock secret for testing"
regex = '''secret'''
"""

dex_proc = None


def validate_token(token):
    try:
        req = Request(f"{DEX_ISSUER}/userinfo")
        req.add_header("Authorization", f"Bearer {token}")
        urlopen(req)
        return True
    except (HTTPError, URLError):
        return False


class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        auth = self.headers.get("Authorization", "")
        if not auth.startswith("Bearer ") or not validate_token(auth[7:]):
            self.send_response(401)
            self.send_header("WWW-Authenticate", f'Bearer realm="{DEX_ISSUER}"')
            self.end_headers()
            return

        if self.path.startswith("/patterns/gitleaks/"):
            self.send_response(200)
            self.send_header("Content-Type", "text/plain")
            self.end_headers()
            self.wfile.write(GITLEAKS_CONFIG.encode())
            return

        self.send_response(200)
        self.end_headers()
        self.wfile.write(b"authenticated\n")

    def log_message(self, format, *args):
        print(f"[pattern-server] {args[0]}", file=sys.stderr)


def wait_for_dex():
    for _ in range(30):
        try:
            urlopen(f"{DEX_ISSUER}/.well-known/openid-configuration")
            return
        except (HTTPError, URLError):
            time.sleep(0.5)
    print("error: dex did not start in time", file=sys.stderr)
    sys.exit(1)


def cleanup(*_):
    global dex_proc
    if dex_proc and dex_proc.poll() is None:
        dex_proc.terminate()
        dex_proc.wait()
    sys.exit(0)


def main():
    global dex_proc

    if not shutil.which("dex"):
        print("error: 'dex' not found in PATH", file=sys.stderr)
        print("install: https://dexidp.io/docs/getting-started/", file=sys.stderr)
        sys.exit(1)

    signal.signal(signal.SIGINT, cleanup)
    signal.signal(signal.SIGTERM, cleanup)

    dex_proc = subprocess.Popen(
        ["dex", "serve", "/dev/stdin"],
        stdin=subprocess.PIPE,
        stdout=sys.stderr,
        stderr=sys.stderr,
    )
    dex_proc.stdin.write(DEX_CONFIG.encode())
    dex_proc.stdin.close()

    wait_for_dex()

    server = HTTPServer((SERVER_HOST, SERVER_PORT), Handler)
    print(f"mock pattern server: http://{SERVER_HOST}:{SERVER_PORT}", file=sys.stderr)
    print(f"dex issuer:          {DEX_ISSUER}", file=sys.stderr)
    print(f"login:               admin@example.com / password", file=sys.stderr)

    try:
        server.serve_forever()
    finally:
        cleanup()


if __name__ == "__main__":
    main()
```

cmd/cmd.go
sharman-rh: Good point, I'll remove the redundancies

cmd/testserver/main.go
sharman-rh: The mock server only works as long as I remove the 
``` 
 redirectURIs:
  - http://127.0.0.1/callback
```
part. Is that ok?

PR #305, opened 2026-08-28
Allow scans to use sources for auth

bplaxco:
# Overview

Provide a way for sources to be leveraged by other parts of leaktk for authentication. This allows things like direct URL scanning against a Jira URL, etc based off of your configured sources.

For now I'm leaving out more generic auth but wanted to start with this since it addresses a current need.

Also this moves a few modules under internal. This is part of the ongoing effort to move things out of pkg and then intentionally provide a public API. Just doing it a bit at a time as I make other changes.

## Change summary

- Provide `Sources.SetHeader(req *http.Request)` to set auth headers for the JSON/URL resources
- Expand RateLimit to be target host aware and provide a single instance of it similar to how client is done
- Break `RateLimit.Limit` into `RateLimit.Wait` and `RateLimit.Update` to work better for the other cases
- Move some packages to `internal`
- Rename custom `http` package to `httpclient` to avoid clashes with `net/http`
- Update/improve docs around config/sources/collect.

For the rate limits, I updated `url_source.go` `json_source.go`, and `atlassian.go` to add a for loop that looks roughly like:

```go
var (
    err  error
    req  *http.Request
    resp *http.Response
)

for resp == nil || resp.StatusCode == http.StatusTooManyRequests {
    req, err = http.NewRequestWithContext(ctx, method, url, body)
    if err != nil {
        // handle errors...
    }

    // Let the sources set any applicable auth headers
    if err := s.Sources.SetHeader(req); err != nil {
        // handle errors...
    }

    // Wait based on the current state of the rate limit
    if err := s.RateLimit.Wait(req); err != nil {
        // handle errors...
    }

    // do the request and get the response
    // resp, err = ....

    s.RateLimit.Update(resp)
}
```

It tries again when there's too many requests and `RateLimit` is in charge of backing off based on what's set in the response.

# To-Do

- [x] Integrate Sources into the scan where they can be leveraged by the URL/JSON betterleaks "sources" (aka resources). 
- [x] QA testing
- [x] Write docs


docs/collect.md
```diff
@@ -7,7 +7,7 @@ for gathering information needed for remediation and analysis.
 
 - `AtlassianCloudAdmin`
 
-Planned: GitHubOrganization GitHubEnterpriseAccount, LDAP, TBD...
+Planned: GitHub GitLab, LDAP, URL, TBD...
```
thewizzy, line 10:
```suggestion
Planned: GitHub, GitLab, LDAP, URL, TBD...
```

thewizzy (approved):
LGTM. 
Really liked the Rate Limiter!

docs/collect.md
bplaxco: oop saw the LTGM and merged. I'll fix that as well. good catch.

bplaxco: https://github.com/leaktk/leaktk/pull/308

